PR #3, opened 2019-07-21
WIP: Refactor transport layer to support SDIO

MabezDev:
This PR begins to separate the transport from the SD protocol, so we can use other interfaces such as SDIO. Before I went any further, I thought I'd better check that this is something you are interested in, and to see if you have any comments on how to approach it.

### DIFF
--- a/src/sdmmc/mod.rs
+++ b/src/sdmmc/mod.rs
@@ -1,0 +1,148 @@
+
+
+mod sdio;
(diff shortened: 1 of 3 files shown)


thejpster:
I'm going to close this, because we didn't get it out of WIP and merged before a whole bunch of stuff happened. Feel free to rebase on current head and open a new one.

PR #6, opened 2019-10-23
Fix erroneous bitshift for crc check

### LINKED ISSUE #5: Bug: crash on right shift overflow
The following CRC16 right shift is incorrect; should only shift by 8 bits. Will issue PR. I have no clue as to why this was not an issue earlier.

https://github.com/thejpster/embedded-sdmmc-rs/blob/ef05517d030ea6672fcb6874fc88a91e7c706bab/src/sdmmc.rs#L352

Wassasin:
Fixes #5

### DIFF
--- a/src/sdmmc.rs
+++ b/src/sdmmc.rs
@@ -349,7 +349,7 @@ where
         for &b in buffer.iter() {
             self.send(b)?;
         }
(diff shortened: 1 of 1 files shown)


thejpster:
Looks legit. Thanks!

PR #9, opened 2019-11-06
Fixed bug when allocating cluster when current_cluster is not yet at end

Wassasin:
Steps to reproduce:

1. Have file of 16384 bytes (32 clusters)
2. Open file with ReadWriteCreateOrAppend
3. Write to file.

Because opening the file and `file.seek_from_end(0)` does not update `file.current_cluster`, `alloc_cluster` will alloc a new cluster and attach it as the next cluster of the first cluster. This would effectively delete all the file contents following the first cluster.

Changed `find_data_on_disk` to actually seek to the next file cluster by using &mut, even when throwing errors. (Such as EndOfFile). I clone the current cluster because technically this gets updated for the file object after effectively writing the new block.

If desired I could provide with an `.img` file for this test case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,3 +14,7 @@ readme = "README.md"
 embedded-hal = "0.2.2"
 byteorder = { version = "1", default-features = false }
 nb = "0.1"
+log = "0.4"
+
+[dev-dependencies]
+env_logger = "0.7"
--- a/examples/create_test.rs
+++ b/examples/create_test.rs
@@ -110,6 +110,7 @@ impl TimeSource for Clock {
 }
 
 fn main() {
+    env_logger::init();
     let mut args = std::env::args().skip(1);
     let filename = args.next().unwrap_or("/dev/mmcblk0".into());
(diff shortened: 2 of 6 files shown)


thejpster:
Would it be possible to write a unit test which demonstrates this bug, and that this fix does actually fix it?

Wassasin:
Sure, I might be able to write it next week. I'll keep you posted.

thalesfragoso:
I haven't done a practical test, but I took a look at the code and indeed the logic is flawed as explained, this is certainly a bug and it also can lead to FAT corruption. This fix seems to address the problem correctly. 
Thanks.

thejpster:
If you see my `bug-write-cluster-alloc` branch, I've update the `write_test` example to append a larger block to the file, and it seems to hang indefinitely (both on master, and with your fix). Any ideas what's wrong?

thejpster:
Fixed it. There was a bug finding free clusters in the FAT.

thejpster:
Thank you!

PR #16, opened 2019-12-16
add partition id used by macosx

jacobrosenthal:
Im seeing this partition id when formatted with disktuil gui or cli from a mac AND when formatted with the official sd formatter from on a mac.

I couldnt find any combination that would use something different so I think it needs to be supported.


thejpster:
You'd think with a CHS partition type we should use the CHS geometry, but obviously that doesn't make sense with an SD card. I'm happy to add this for now, and if I ever build that SPI to IDE interface (or if Phil adds file system support to his x64 OS), I'll work out how to tell if CHS needs special support.

PR #22, opened 2020-08-15
Refactor transport layer to support SDIO

mattico:
I've updated #3 with the latest code.

From the presence of `sdio.rs` it seems the intention was to provide an implementation of the SDIO transport. I don't think that's practical at the moment. There isn't an interface to SDIO in `embedded_hal` and likely won't be for a long time. Therefore an implementation of the SDIO transport would have to be tied to a device-specific HAL like `stm32h7xx_hal`. We could add such an implementation behind a feature flag, but device HALs often have their own feature flags for specific device support which we would also have to expose and now it starts to get messy. I think it would be better to expect such implementations to live in user code or in external crates.

I do intend to implement the SDIO transport for STM32H7. If you think it makes sense I can add that as a separate crate in this repo.


mattico:
Turns out the STM32H7 HAL already provides `read_block` and `write_block` functions (while not exposing the cmd functions). This `Transport` interface might be useful for sharing the protocol implementation among several HALs but I'm just going to implement `BlockDevice` on top of that.

thejpster:
Apologies for the delay! This looks really good. My only question is whether we should have a `read_data` impl for SPI transport which uses the blocking SPI functions to transfer data in bigger chunks. This would be more efficient on something like an nRF52 which sets up a DMA transaction for each SPI transfer, and so transferring all 512 bytes would be much more efficient than 512 transfers of 1 byte.

thejpster:
Also, is SDIO the right term for the high-speed SDMMC interface? I thought SDIO for specifically about using the SD card interface to drive I/O devices, like Wi-Fi cards. Wikipedia suggests the SD transports are:

* SPI bus mode
* One-bit SD bus mode
* Four-bit SD bus mode
* Two differential lines SD UHS-II mode

mattico:
Yeah, it seems a bit silly to use the single-byte SPI methods looking at this now.

SDIO does seem like a not-ideal name due to the existence of SD I/O as you say but 1. I've never seen or heard of an SD I/O device in real life outside of wikipedia. 2. From what I've seen everyone calls the SD bus SDIO and I couldn't think of a better name. I guess we could just call it "Parallel"? `SdMmcParallel`

thejpster:
SDIO is definitely found in the wild. Here are a bunch of modules you can buy from Mouser, for example:

https://www.mouser.co.uk/Embedded-Solutions/Wireless-RF-Modules/WiFi-80211-Modules/_/N-6l7qa?P=1z0jdef

I think I'm fine with just having the Transport trait here and letting, say, the stm32-rs HALs do the implementation, named after the STM32 Peripheral they use. Without an embedded-hal trait for an SDMMC Peripheral, or putting a bunch of chip specific code in here, I don't think there's much else we can do. If we have to, we can refer to SD Bus Mode and SPI Mode.

mattico:
Alright I used the blocking API for the most important and obvious cases. I don't actually use the SPI API and can't test it so I'd like to leave it at that for now.

thejpster:
I'm going to close this, because we didn't get it merged before a whole bunch of stuff happened. Sorry about that. Feel free to rebase on current head and open a new one.

PR #23, opened 2020-10-10
Add support for file deletion

glbsalazar:
Add support for file deletion, also added an accompanying test for the file deletion.
Added disk.img to .gitignore to prevent accidental commits


src/fat.rs
```diff
@@ -1003,6 +1003,117 @@ impl FatVolume {
...
+                            let start = entry * OnDiskDirEntry::LEN;
+                            let end = (entry + 1) * OnDiskDirEntry::LEN;
+                            let dir_entry = OnDiskDirEntry::new(&blocks[0][start..end]);
+                            if dir_entry.is_end() {
+                                // Can quit early
+                                return Err(Error::FileNotFound);
+                            } else if dir_entry.matches(&match_name) {
+                                // let mut blockWrite = &blocks[0][start..end];
+                                // blockWrite[0] = 0xE5;
+                                let mut blocks = [Block::new()];
```
thejpster, line 1050: This looks like it destroys a bunch of directory entries, instead of modifying the leading byte in the specific directory entry we want to scratch. Is this correct?

src/fat.rs
```diff
@@ -1003,6 +1003,117 @@ impl FatVolume {
...
+            }
+            FatSpecificInfo::Fat32(fat32_info) => {
+                let match_name =
+                    ShortFileName::create_from_str(name).map_err(Error::FilenameError)?;
+                let mut current_cluster = match dir.cluster {
+                    Cluster::ROOT_DIR => Some(fat32_info.first_root_dir_cluster),
+                    _ => Some(dir.cluster),
+                };
+                let mut blocks = [Block::new()];
+                while let Some(cluster) = current_cluster {
```
thejpster, line 1082: This loops seems very similar to the FAT16 version. Can we break it out into its own function, passing in the relevant values from `fat16_info` or `fat32_info` as arguments, so we only have to write it once?

src/lib.rs
```diff
@@ -585,6 +589,42 @@ where
...
+    ) -> Result<(), Error<D::Error>> {
+        debug!(
+            "delete_file(volume={:?}, dir={:?}, filename={:?}",
+            volume, dir, name
+        );
+        let dir_entry = match &volume.volume_type {
+            VolumeType::Fat(fat) => fat.find_directory_entry(self, dir, name),
+        };
+
+        let dir_entry = match dir_entry {
```
thejpster, line 607: Can we shorten this to `let dir_entry = dir_entry.map_err(Error::FileNotFound)?`;

thejpster (changes requested):
Thanks for this PR! Looks great to me in general, but I had one correctness query and a couple of comments about whether we can reduce code duplication.

src/lib.rs
glbsalazar: I shorten it even further since I can directly evaluate it from the previous match

src/fat.rs
glbsalazar: Thanks for the feedback, it was not intended. It should be similar to the fat32 case.

src/fat.rs
glbsalazar: I reduced some code duplication in both the delete and the find methods. Please check if it was similar to what you intended

glbsalazar:
@thejpster Thank you for the quick review! I made some changes aligned with your comments.
Regarding the code duplication I tried reducing it without sacrificing too much the readability.

src/fat.rs
```diff
@@ -964,33 +949,120 @@ impl FatVolume {
...
+            let end = (entry + 1) * OnDiskDirEntry::LEN;
+            let dir_entry = OnDiskDirEntry::new(&blocks[0][start..end]);
+            if dir_entry.is_end() {
+                // Can quit early
+                return Err(Error::FileNotFound);
+            } else if dir_entry.matches(&match_name) {
+                // Found it
+                // Safe, since Block::LEN always fits on a u32
+                let start = u32::try_from(start).unwrap();
+                return Ok(dir_entry.get_entry(FatType::Fat16, block, start));
```
thalesfragoso, line 1001: This seems to be a bug on the old code where it uses `FatType::Fat16` for both FAT16 and FAT32.

glbsalazar: Thanks for the input! I'll take a look at it, but I tested it with Fat16 and Fat32 and did not see any differences.
I'll take a look at it later tonight

glbsalazar: Fixed it. Now the proper FatType is being used for Fat32 entries

thejpster (approved):
Apologies for the delay

PR #26, opened 2020-10-17
Make some changes to improve readability

qiuchengxuan:


src/fat.rs
```diff
@@ -353,47 +356,22 @@ impl<'a> OnDiskDirEntry<'a> {
...
-                    .unwrap();
-            buffer[12] =
-                core::char::from_u32(u32::from(LittleEndian::read_u16(&self.data[30..=31])))
-                    .unwrap();
-            Some((is_start, sequence, buffer))
+            let mut buffer = [0u8; 26];
+            buffer[..10].copy_from_slice(&self.data[1..=10]);
+            buffer[10..22].copy_from_slice(&self.data[14..=25]);
+            buffer[22..].copy_from_slice(&self.data[28..=31]);
+            let data: &[u16; 13] = unsafe { core::mem::transmute(&buffer) };
```
thalesfragoso, line 367: Couldn't this be unaligned ?

src/fat.rs
```diff
@@ -353,47 +356,22 @@ impl<'a> OnDiskDirEntry<'a> {
...
+            let mut string: String<U26> = String::new();
+            // Little-endian UCS-2 to Big-endian UTF-8
+            for i in 0..data.len() {
+                let ch = unsafe { core::char::from_u32_unchecked(u16::from_le(data[i]) as u32) };
```
thalesfragoso, line 371: Could you specify in a comment why do you think this is sound ? Same thing for all the others `unsafe` uses.

src/fat.rs
```diff
@@ -519,10 +497,8 @@ impl FatVolume {
                     Cluster::END_OF_FILE => 0xFFFF,
                     _ => new_value.0 as u16,
                 };
-                LittleEndian::write_u16(
-                    &mut blocks[0][this_fat_ent_offset..=this_fat_ent_offset + 1],
-                    entry,
-                );
+                let p_entry: *mut u16 = &mut blocks[0][this_fat_ent_offset] as *mut _ as *mut u16;
+                unsafe { *p_entry = u16::to_le(entry) };
```
thalesfragoso, line 501: Couldn't this be unaligned ?

src/fat.rs
```diff
@@ -539,14 +515,9 @@ impl FatVolume {
...
-                let existing = LittleEndian::read_u32(
-                    &blocks[0][this_fat_ent_offset..=this_fat_ent_offset + 3],
-                );
-                let new = (existing & 0xF000_0000) | (entry & 0x0FFF_FFFF);
-                LittleEndian::write_u32(
-                    &mut blocks[0][this_fat_ent_offset..=this_fat_ent_offset + 3],
-                    new,
-                );
+                let p_offset = &mut blocks[0][this_fat_ent_offset] as *mut _ as *mut u32;
+                let existing = u32::from_le(unsafe { *p_offset });
```
thalesfragoso, line 519: Unaligned ?

src/fat.rs
```diff
@@ -576,22 +547,14 @@ impl FatVolume {
...
-                    0xFFF8..=0xFFFF => {
-                        // There is no next cluster
-                        Err(Error::EndOfFile)
-                    }
-                    f => {
-                        // Seems legit
-                        Ok(Cluster(u32::from(f)))
-                    }
+                let p_fat_entry = &blocks[0][this_fat_ent_offset] as *const _ as *const u16;
+                match u16::from_le(unsafe { *p_fat_entry }) {
```
thalesfragoso, line 551: Unaligned ?

src/fat.rs
```diff
@@ -602,26 +565,16 @@ impl FatVolume {
...
-                    0x0000_0001 | 0x0FFF_FFF8..=0x0FFF_FFFF => {
-                        // There is no next cluster
-                        Err(Error::EndOfFile)
-                    }
-                    f => {
-                        // Seems legit
-                        Ok(Cluster(f))
-                    }
+                let p_fat_entry = &blocks[0][this_fat_ent_offset] as *const _ as *const u32;
+                match u32::from_le(unsafe { *p_fat_entry }) & 0x0FFF_FFFF {
```
thalesfragoso, line 569: Unaligned ?

src/fat.rs
```diff
@@ -1038,10 +991,8 @@ impl FatVolume {
                         .map_err(Error::DeviceError)?;
 
                     while this_fat_ent_offset <= Block::LEN - 2 {
-                        let fat_entry = LittleEndian::read_u16(
-                            &blocks[0][this_fat_ent_offset..=this_fat_ent_offset + 1],
-                        );
-                        if fat_entry == 0 {
+                        let p_fat_entry = &blocks[0][this_fat_ent_offset] as *const _ as *const u16;
+                        if u16::from_le(unsafe { *p_fat_entry }) == 0 {
```
thalesfragoso, line 995: Unaligned ?

src/fat.rs
```diff
@@ -1070,10 +1021,8 @@ impl FatVolume {
                         .map_err(Error::DeviceError)?;
 
                     while this_fat_ent_offset <= Block::LEN - 4 {
-                        let fat_entry = LittleEndian::read_u32(
-                            &blocks[0][this_fat_ent_offset..=this_fat_ent_offset + 3],
-                        ) & 0x0FFF_FFFF;
-                        if fat_entry == 0 {
+                        let p_fat_entry = &blocks[0][this_fat_ent_offset] as *const _ as *const u32;
+                        if u32::from_le(unsafe { *p_fat_entry }) & 0x0FFF_FFFF == 0 {
```
thalesfragoso, line 1025: Unaligned ?

src/lib.rs
```diff
@@ -276,36 +275,25 @@ where
...
-            let lba_start = LittleEndian::read_u32(
-                &partition[PARTITION_INFO_LBA_START_INDEX..(PARTITION_INFO_LBA_START_INDEX + 4)],
-            );
-            let num_blocks = LittleEndian::read_u32(
-                &partition[PARTITION_INFO_NUM_BLOCKS_INDEX..(PARTITION_INFO_NUM_BLOCKS_INDEX + 4)],
-            );
+            let pointer = &partition[PARTITION_INFO_LBA_START_INDEX] as *const _ as *const u32;
+            let lba_start = u32::from_le(unsafe { *pointer });
+            let pointer = &partition[PARTITION_INFO_NUM_BLOCKS_INDEX] as *const _ as *const u32;
+            let num_blocks = u32::from_le(unsafe { *pointer });
```
thalesfragoso, line 296: Unaligned ?

thalesfragoso (changes requested):
Thanks for the PR, I left some comments.

src/fat.rs
qiuchengxuan: Compiler will ensure it's aligned by word

src/fat.rs
qiuchengxuan: Modified a bit and explained by variable name. The original code unwrapped the result from `core::char::from_u32`, which leads to panic if it's not a valid rust char, Silently ignore invalid char would be a better choice, otherwise returns Result type for caller to resolve should be another choice.

src/fat.rs
qiuchengxuan: this_fat_ent_offset is 2B aligned

src/fat.rs
qiuchengxuan: same as above

src/fat.rs
qiuchengxuan: same as above

src/fat.rs
qiuchengxuan: 4B aligned

src/fat.rs
qiuchengxuan: 2B aligned

src/fat.rs
qiuchengxuan: same as above

src/lib.rs
qiuchengxuan: same as above

src/fat.rs
thalesfragoso: Creating an invalid `char` is UB, which IMO is way worse than a panic.

src/fat.rs
thalesfragoso: Could you explain why you think this is the case ? `buffer` has an alignment of 1, while for data that is 2, why do you think that can't cause an unaligned reference ?

src/fat.rs
thalesfragoso: `Block` is an abstraction over an array of u8s, which has an alignment of 1, no ?

src/fat.rs
qiuchengxuan: A panic generated by third-party library is also unacceptable, as a third-party library you should not make choice for the user, but give the choise to user, or aware user filename is possibly invalid.
Even if a filename contains invalid char, it's still part of the filename, and we just don't know it's encoding.

thalesfragoso: Then we should return an error, instead of causing undefined behavior.

src/fat.rs
qiuchengxuan: look at the assembly code on x86-64 release object
```
0000000000013140 <_ZN14embedded_sdmmc3fat14OnDiskDirEntry12lfn_contents17h5e14fbe2a1624622E>:
   13140:       48 83 ec 78             sub    $0x78,%rsp
   13144:       48 89 f1                mov    %rsi,%rcx
   13147:       48 8b 76 08             mov    0x8(%rsi),%rsi
   1314b:       48 83 fe 0b             cmp    $0xb,%rsi
   1314f:       0f 86 ad 01 00 00       jbe    13302 <_ZN14embedded_sdmmc3fat14OnDiskDirEntry12lfn_contents17h5e14fbe2a1624622E+0x1c2>                                                        
   13155:       48 89 f8                mov    %rdi,%rax
   13158:       48 8b 11                mov    (%rcx),%rdx
   1315b:       8a 4a 0b                mov    0xb(%rdx),%cl
   1315e:       80 e1 0f                and    $0xf,%cl
   13161:       80 f9 0f                cmp    $0xf,%cl
   13164:       0f 85 3a 01 00 00       jne    132a4 <_ZN14embedded_sdmmc3fat14OnDiskDirEntry12lfn_contents17h5e14fbe2a1624622E+0x164>                                                        
   1316a:       44 8a 0a                mov    (%rdx),%r9b
   1316d:       0f 57 c0                xorps  %xmm0,%xmm0
   13170:       0f 11 44 24 3a          movups %xmm0,0x3a(%rsp)
   13175:       48 8b 4a 01             mov    0x1(%rdx),%rcx
   13179:       48 89 4c 24 30          mov    %rcx,0x30(%rsp)
   1317e:       0f b7 4a 09             movzwl 0x9(%rdx),%ecx
   13182:       66 89 4c 24 38          mov    %cx,0x38(%rsp)
   13187:       48 83 fe 19             cmp    $0x19,%rsi
```

It's not hard to figure out the buffer is starting from 0x30(%rsp)，which is aligned by word

src/fat.rs
qiuchengxuan: Changed to if not valid char, treat as no LFN.

src/fat.rs
qiuchengxuan: changed to definition of [u16; 13] to ensure at least 2B alignment

src/fat.rs
qiuchengxuan: reverted

src/fat.rs
qiuchengxuan: reverted

src/fat.rs
qiuchengxuan: reverted

src/fat.rs
qiuchengxuan: reverted

src/lib.rs
qiuchengxuan: reverted

src/fat.rs
qiuchengxuan: reverted

src/fat.rs
```diff
@@ -353,50 +357,27 @@ impl<'a> OnDiskDirEntry<'a> {
...
-        } else {
-            None
+    pub fn lfn_contents(&self) -> Option<(bool, u8, String<U26>)> {
+        if !self.is_lfn() {
+            return None;
         }
+        let is_start = (self.data[0] & 0x40) != 0;
+        let sequence = self.data[0] & 0x1F;
+        let mut buffer = [0u16; 13];
+        let bytes: &mut [u8; 26] = unsafe { core::mem::transmute(&mut buffer) };
```
thejpster, line 367: Is this allowed? I thought multiple mut references was UB.

thejpster: I think this also assumes the endianess of the CPU matches the on disk format, which does not always hold.

thejpster (changes requested):

src/fat.rs
thalesfragoso: Hmm, where is the other reference ? I think you could get into UB if you used `bytes`, then `buffer` and then `bytes` again, but I don't think this is the case. However, I think that we should at least put these operations inside a block `{}` to avoid future problems, `transmute` can be nasty. Anyways, I might also be missing some thing that would make this UB.

I agree about the endianess problem.

thalesfragoso: Hmm, if `transmute` is tying the lifetime of bytes to the lifetime of the passed reference `&mut buffer` then this is UB. I guess we should avoid `transmute` here, we could probably to the same thing with `as` and raw pointers.

thejpster: Or just use the `from_bytes` and `to_bytes` functions and allow the compiler to optimise it all. I don't think there's a need for the fancy tricks here.

thalesfragoso: > I don't think there's a need for the fancy tricks here.

I agree.

thejpster:
I'm going to close this, because we didn't get it out of WIP and merged before a whole bunch of stuff happened. Feel free to rebase on current head and open a new one.

PR #29, opened 2020-11-19
Use blocking spi traits

### LINKED ISSUE #28: Consider switching to `embedded_hal::blocking::spi` traits
_Disclaimer_: I'm not actively using this crate right now.  I was made aware of this potential problem by @Maldus512 (in Rahix/shared-bus#18) and wanted to start a discussion about it here.

Currently, the [`SdMmcSpi`] type is based on the [`FullDuplex`] trait from _embedded-hal_.  The contract for [`FullDuplex`] has requirements on the order of calls to its methods which makes it impossible to allow sound bus sharing in the way [`shared-bus`] works.  Quoting my comment in Rahix/shared-bus#18 with the reason why:

> [...] the `FullDuplex` trait unfortunately has an API that cannot be safely shared in the way `shared-bus` solves sharing.  The reason is that there is an implicit dependency between calls to `send()` and `read()`  which could be broken if multiple bus users interweave these calls.  For example:
> 
> ```text
>     User 1           User 2
>       |                |
>     send()             |
>       |                |
>       |              send() - - # Now received byte from first send is discarded (or worse)
>       |                |
>     read()  - - - - - - - - - - # Reads data from second write
>                        |
>                      read() - - # HAL-API violation!!
> ```

To allow bus sharing with [`SdMmcSpi`], I would suggest switching `embedded-sdmmc` to use the traits from [`embedded_hal::blocking::spi`] instead as they do not suffer this problem and can be used for bus sharing more safely (though still not concurrently across multiple execution contexts[¹]).

If I am not mistaken, the only place where this change would need to happen would be this method:

https://github.com/rust-embedded-community/embedded-sdmmc-rs/blob/6071df4fee21175cec9e664321c828bddc60dacf/src/sdmmc.rs#L417-L422

By using [`embedded_hal::blocking::spi::Transfer`] instead, I think it could just look like this:

```rust
    /// Send one byte and receive one byte.
    fn transfer(&self, out: u8) -> Result<u8, Error> {
        let mut spi = self.spi.borrow_mut();
        spi.transfer(&mut [out])
            .map(|b| *b[0])
            .map_err(|_e| Error::Transport)
    }
```

(Of course, by doing a larger change, this could be done more efficiently).

[`SdMmcSpi`]: https://docs.rs/embedded-sdmmc/0.3.0/embedded_sdmmc/struct.SdMmcSpi.html
[`FullDuplex`]: https://docs.rs/embedded-hal/0.2.4/embedded_hal/spi/trait.FullDuplex.html
[`shared-bus`]: https://github.com/Rahix/shared-bus
[`embedded_hal::blocking::spi`]: https://docs.rs/embedded-hal/0.2.4/embedded_hal/blocking/spi/index.html
[¹]: https://github.com/Rahix/shared-bus/issues/8
[`embedded_hal::blocking::spi::Transfer`]: https://docs.rs/embedded-hal/0.2.4/embedded_hal/blocking/spi/trait.Transfer.html

eldruin:
- Shift to using blocking SPI traits for peripheral sharing ease
- Fix updating `embedded-hal` version to version 0.2.3 where the `digital::v2` traits were actually added.
- Update `env_logger` dependency
- Add Changelog

Thanks to @Rahix and @Maldus512
Fixes #28

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,15 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Changes
+- [breaking-change] Use SPI blocking traits instead to ease SPI peripheral sharing.
+  See: https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/28
+
+
+[Unreleased]: https://github.com/rust-embedded-community/embedded-sdmmc-rs/compare/v0.3.0...develop
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,5 +17,5 @@ nb = "0.1"
 log = "0.4"
 
 [dev-dependencies]
-env_logger = "0.7"
 hex-literal = "0.3"
+env_logger = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,9 +11,8 @@ edition = "2018"
 readme = "README.md"
 
 [dependencies]
-embedded-hal = "^0.2.2"
+embedded-hal = "0.2.3"
 byteorder = { version = "1", default-features = false }
-nb = "0.1"
 log = "0.4"
 
 [dev-dependencies]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,10 +17,9 @@
 //! # struct DummyCsPin;
 //! # struct DummyUart;
 //! # struct DummyTimeSource;
-//! # impl embedded_hal::spi::FullDuplex<u8> for  DummySpi {
+//! # impl embedded_hal::blocking::spi::Transfer<u8> for  DummySpi {
 //! #   type Error = ();
-//! #   fn read(&mut self) -> nb::Result<u8, ()> { Ok(0) }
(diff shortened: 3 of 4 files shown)


Rahix (approved):
Looks good to me at least :)

thejpster:
I only looked at this quickly, but I had a question around the chip select. Is there a risk the card will be unselected in the middle of a transaction when using a shared bus object? And if so, does the card care?

eldruin:
> I only looked at this quickly, but I had a question around the chip select. Is there a risk the card will be unselected in the middle of a transaction when using a shared bus object? And if so, does the card care?

shared-bus allows sharing when the objects are not already borrowed so taking away the spi peripheral in the middle of a transaction is not possible.
What could happen, though, is that an interrupt arrives between the time of changing the CS pin and the use of the SPI peripheral. If in that interrupt the SPI peripheral is borrowed, a situation would come up where multiple peripherals are selected.
This is a known limitation of shared-bus and there is a proposal around a managed CS trait to solve it. See: https://github.com/Rahix/shared-bus/issues/8
Other than that I believe there is no danger being added here. Even though the `FullDuplex` trait is being used, it is being used in a blocking way already.

Rahix:
> If in that interrupt the SPI peripheral is borrowed, a situation would come up where multiple peripherals are selected.

That is currently impossible because the bus proxy for SPI is `!Send`.  So you'd need some unsafe code to do this.

> This is a known limitation of shared-bus and there is a proposal around a managed CS trait to solve it.

Also see <https://github.com/rust-embedded/embedded-hal/pull/245>.

thejpster (approved):

PR #32, opened 2021-02-07
Fix init sequence (CMD0).

andreyk0:
Avoid wait_not_busy() calls before SD card is reset, the first command
after an idle state needs to be CMD0 to reliably reset an SD card.


andreyk0:
While this seems to fix it on one type of micro a different init problem appears on another. There must be some other issue in my init sequence.

PR #45, opened 2022-01-12
optionally use defmt + add defmt::Format to required structs

gauteh:
Would this be of interest?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -3,25 +3,34 @@ name: Rust
 on: [push, pull_request]
 
 jobs:
-  build:
+  formatting:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
+    - name: Check formatting
+      run: cargo fmt -- --check
+
+  build-test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        features: ['log', 'defmt-log']
+    steps:
+    - uses: actions/checkout@v1
     - name: Build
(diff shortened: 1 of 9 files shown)


jonathanpallant (approved):

eldruin (commented):
Looks good to me. Could you add an entry to the changelog?

gauteh:
Changelog entry added.

eldruin:
There are some problems building this. Could you have a look?
Could you also adapt the CI to test this feature as well and also document this feature both in the README and in the `lib.rs`?

gauteh:
Had to make some minor changes, please re-run CI.

eldruin:
There is an error in the script preventing execution:
```
Error : .github#L1
every step must define a `uses` or `run` key
```

eldruin (commented):
Looks good! One last thing before the merge, though: Could you also improve the CI to run without any features i.e. just `--no-default-flags`?
One easy way to do this is to change the `features` values to `['', '--features log', '--features defmt-log']` and adapt the rest accordingly

gauteh:
> Looks good! One last thing before the merge, though: Could you also improve the CI to run without any features i.e. just `--no-default-flags`? One easy way to do this is to change the `features` values to `['', '--features log', '--features defmt-log']` and adapt the rest accordingly

That won't work as it is now, you have to select a logger.

eldruin (approved):
Interesting. Ok, then let's merge this as-is.
Thank you for your work!

PR #49, opened 2022-02-20
Let users specify Controller handler capacity

olback:


jannic:
As I understand it, this is more or less the same as the (newer, but already merged) pull request https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/59. Can this be closed, or is PR contain additional improvements which should still be merged?

olback:
Looks like the same thing.

PR #51, opened 2022-04-17
Let sdmmc::Error derive defmt::Format (if enabled)

ithinuel:
This saves users from having to used `Debug2Format` and benefit from defmt for printing those errors too.


eldruin:
Superseeded by #54. Still, thank you!

PR #58, opened 2022-07-10
Fix env_logger dependency enabling std feature in log

### LINKED ISSUE #44: log pulls in std: how do I disable the std feature on the log crate?
Optionally add `default-features = false` to log in you Cargo.toml.

Is this even possible?

```
embedded-sdmmc-rs on  develop [$] via 🦀 v1.59.0-nightly
❯ cargo build --target=thumbv7em-none-eabihf
   Compiling log v0.4.14
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv7em-none-eabihf` target may not support the standard library
  = note: `std` is required by `log` because it does not declare `#![no_std]`
  = help: consider building the standard library from source with `cargo build -Zbuild-std`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `log` due to previous error
```

eldruin:
One last fix before the 0.4.0 release

Fixes #44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added `Controller::has_open_handles` and `Controller::free` methods.
 - [breaking-change] Changed interface to enforce correct SD state at compile time.
 - [breaking-change] Added custom error type for `File` operations.
+- Fix `env_logger` pulling in the `std` feature in `log` in library builds.
+- Raise the minimum supported Rust version to 1.56.0.
 - Code tidy-ups and more documentation.
 
 ## [Version 0.3.0](https://github.com/rust-embedded-community/embedded-sdmmc-rs/releases/tag/v0.3.0)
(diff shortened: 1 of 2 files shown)


thejpster:
The only change to Cargo.toml is `resolver = "2"`. Is that right?

eldruin:
Yes, that is correct. With the new resolver the feature is not enabled.

thejpster (approved):

thejpster:
So you could equally fix this by specifying an MSRV high enough that rustc has `resolver="2"` on by default? And this change instead sets MSRC high enough that `resolver` is an understood attribute?

eldruin:
Kind of, yes. `resolver="2"` became the default when using edition 2021 (Rust >= 1.56.0).

eldruin:
Ok I have changed this now to use edition 2021 instead, which has resolver="2" by default and noted the MSRV update to 1.56.0, which [lots of people have already](https://lib.rs/stats#rustc), in the changelog. 
We do not have an explicit MSRV here, though.

thejpster (approved):

PR #59, opened 2022-07-26
Customisable maximum number of open directories and files

ostenning:
Within my embedded app I require multiple data streams to multiple files, greater than the default supported amount which is currently `4`.

This PR simply allows customization of the maximum allowed open directories and files by replacing the hardcoded constants `MAX_OPEN_DIRS` and `MAX_OPEN_FILES` with const generics.

I've tested this change with stm32s and [this block device](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/sdmmc.rs#L1175) implementation and everything works and I currently have 9 files open simultaneously.

I'm curious to know why the default max was originally set at 4 directories and 4 files. Are there any potential ramifications for increasing these?

Any feedback or assistance with improving this is greatly appreciated.

Kind regards

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 [Unreleased]: https://github.com/rust-embedded-community/embedded-sdmmc-rs/compare/v0.4.0...develop
 
+### Changes
+- Add `MAX_DIRS` and `MAX_FILES` generics to `Controller` to allow an arbitrary numbers of concurrent open directories and files.
+
 ## [Version 0.4.0](https://github.com/rust-embedded-community/embedded-sdmmc-rs/releases/tag/v0.4.0)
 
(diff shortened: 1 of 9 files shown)


eldruin (commented):
Thank you!
Maybe @thejpster has a reason for those numbers?

thejpster:
More open files uses more RAM. No reason for picking 4 though.

eldruin (commented):
Alright. This looks good to me. There is still a failing test, though.
Could you also add some documentation about the maximum number of dirs and files being configurable and how to do that?
Also, could you add an entry to the changelog about this?

ostenning:
> Alright. This looks good to me. There is still a failing test, though. Could you also add some documentation about the maximum number of dirs and files being configurable and how to do that? Also, could you add an entry to the changelog about this?

@eldruin  Sure thing! I've made an update fixing the documentation, updating the README and CHANGLOG

eldruin (approved):
Great! thank you!

Issue #56, opened 2022-06-17
SdMmcSpi::acquire should consume self, not use &mut ref


ceigel:
Currently SdMmcSpi::acquire returns a BlockSpi object which contains a mutable reference to the SdMmcSpi. This is a construct that is very hard to work with. I'm reading parts of a file inside an interrupt, so I have to keep alive both a SdMmcSpi and a BlockSpi. Moreover the SdMmcSpi has to be placed in a static variable so that it is not moved once the BlockSpi is constructed. This is quite hard to use.
I believe it would be easier if the SdMmcSpi would be moved inside BlockSpi.

gauteh:
This would make it hard to re-acquire the BlockDeviceSPI in case of some error. I've wrapped the BlockDeviceSpi in a handle which makes it a bit easier to work with: https://github.com/gauteh/sfy/blob/main/sfy-buoy/src/storage/mod.rs#L204, depends on #63 .

thejpster:
Closed by #80

PR #64, opened 2022-09-26
Make `BlockSpi` take ownership of `SdMmcSpi`

jbeaurivage:
The current `BlockSpi` API makes it quite difficult to pass ownership around because it borrows the underlying `SdMmcSpi`. This is especially unergonomic e.g. in RTIC, where a `Controller` could be initialized and is expected to be returned by the `init` function. See #56.

This PR proposes an alternative where `BlockSpi` takes ownership of the `SdMmcSpi`, while maintaining the typelevel guarantees that the card is properly initialized before a `BlockSpi` can be constructed.

This is achieved by using a `try_init` method on `SdMmcSpi`, which returns a `Result<InitToken, Error>`. This token can then be used in `acquire` to construct a `BlockSpi`. Since a succesful call to `try_init` and `try_init_with_opts` is the only (safe) way to acquire a `InitToken`, a `BlockSpi` is technically guaranteed to hold a correctly initialized card.

Initialization retries is also very easy using this API, because `try_init*` can be called repeatedly without having to reborrow the `SdMmcSpi`:

``` rust
pub fn init_retries<SPI, CS>(mut spi: SdMmcSpi<SPI, CS>) -> BlockSpi<SPI, CS>
where
    SPI: embedded_hal::blocking::spi::Transfer<u8>,
    CS: embedded_hal::digital::v2::OutputPin,
    <SPI as embedded_hal::blocking::spi::Transfer<u8>>::Error: core::fmt::Debug,
{
    let mut maybe_init = spi.try_init();
    loop {
        match maybe_init {
            Ok(t) => return spi.acquire(t),
            Err(_) => maybe_init = spi.try_init(),
        }
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,18 @@ designed for readability and simplicity over performance.
 You will need something that implements the `BlockDevice` trait, which can read and write the 512-byte blocks (or sectors) from your card. If you were to implement this over USB Mass Storage, there's no reason this crate couldn't work with a USB Thumb Drive, but we only supply a `BlockDevice` suitable for reading SD and SDHC cards over SPI.
 
 ```rust
-let mut spi_dev = embedded_sdmmc::SdMmcSpi::new(embedded_sdmmc::SdMmcSpi::new(sdmmc_spi, sdmmc_cs), time_source);
+let mut spi_dev = embedded_sdmmc::SdMmcSpi::new(sdmmc_spi, sdmmc_cs);
+let maybe_init = spi_dev.try_init();
 write!(uart, "Init SD card...").unwrap();
-match spi_dev.acquire() {
-    Ok(block) => {
-        let mut cont = embedded_sdmmc::Controller::new(block, time_source);
+match maybe_init {
+    Ok(token) => {
+        let block_spi = spi_dev.acquire(token);
+        let mut cont: Controller<
+            embedded_sdmmc::BlockSpi<DummySpi, DummyCsPin>,
+            DummyTimeSource,
+            4,
+            4,
+        > = Controller::new(block_spi, time_source);
         write!(uart, "OK!\nCard size...").unwrap();
         match cont.device().card_size_bytes() {
             Ok(size) => writeln!(uart, "{}", size).unwrap(),
@@ -28,7 +35,7 @@ match spi_dev.acquire() {
         }
     }
     Err(e) => writeln!(uart, "{:?}!", e).unwrap(),
-}
+};
 ```
 
 ### Open directories and files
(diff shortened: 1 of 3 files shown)


thejpster:
I'm afraid I don't quite follow what doesn't work with the current API (it's been a year or two since I last used this crate).

Are you say the advantage with this change is that you can stash the token somewhere and call `spi.acquire()` again at your leisure?

jbeaurivage:
Yeah, pretty much. The downside of the current API is that storing a mut reference to the `SdMmcSpi` makes it really hard to share a `BlockSpi` between different contexts (ie, interrupt handlers). 

In my use case, the SD card is initialized in an RTIC `init` handler, then shared between multiple tasks. Therefore the initialization flow has to go like this:

* A BlockSpi has to hold a `&'static mut SdMmcSpi` since RTIC resources can't hold structs that specify non-static lifetimes
* Getting a 'static SdMmcSpi is doable, but definitely not very ergonomic
* Since BlockSpi wants a static reference, if it is dropped (eg, a failed `acquire`), the SdMmcSpi can't be borrowed again. So you have only one chance at calling `acquire`. If the card isn't plugged in yet for example, you won't be able to initialize it again.

This makes it not very friendly for cases where you would want to hot-plug a card when it comes online. The proposed approach also makes it much easier to pass a `BlockSpi/Controller` around, because it's no longer tied to the lifetime of the underlying `SdMmcSpi`. This is definitely desirable in an embedded context, where the driver may be used in different contexts, interrupt handlers, etc. 

In general, I tend to see structs holding references as being much more usable as short-lived constructs. i2c-pio also has the [same](https://docs.rs/i2c-pio/latest/i2c_pio/struct.I2C.html) approach of driver-borrowing-peripheral, and from what I could gather from the RTIC matrix chat, at least one user was struggling with it.

jonathanpallant:
I think I'm falling over the same issue, which I wrote up as #77

thejpster:
I made a bunch of changes. Hopefully the new API works better for you?

jbeaurivage:
@thejpster, I finally found the time to get a prototype up using v0.5. I think it definitely fits the bill better. One thing I think is still missing though: If I try to hotplug a card, the library seems to just loop forever until a card is finally detected. I'd like to be able to handle the failure case where a card isn't plugged in, and try to initialize at a later time. Maybe I'm just not using the API correctly though, but I think having a method that checks whether a card is present and returns a `Result` would be very beneficial. Or an acquire option that specifies the number of retries before giving up. Same with retrying to init the card (when we mark it as uninit).

If you want, I can close this PR as it seems to be obsolete and start a discussion in a new issue.

thejpster:
It's definitely not supposed to loop forever, but I accept the number of retries can be adjusted.

In my application, I detect card removal and insertion, so I only activate the library when I know a card is fitted. However, I did also test by inserting a micro-SD-to-full-SD adapter which did not have an micro-SD card in it, so it fooled the system into thinking a card was inserted when it wasn't, and this took a little while to time out, but it did work.

Issue #33, opened 2021-02-07
SPI init (microcontroller) stuck in TimeoutWaitNotBusy,


andreyk0:
Looked at the init sequence and saw `0xFF` issued before `CMD0`, this [PR](https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/32) fixes the issue for me.

Thanks!

ostenning:
Also facing this issue with an STM32. Trying to use this library with the SPI Block Device. I have confirmed my SPI works fine without this library.

andreyk0:
Hi, I closed PR because I've tried exact same code on a different microcontroller board and there was a different init issue there (that one looked like card wasn't ready to receive CMD0 right after CS was brought low, i.e. affected by this patch).
Then I tried to find more docs / articles about SD and people recommend different things. What's done in this library matches at least some of those recommendations.
In the end I just tried to grab more SD cards out of my junk box and the newer ones seemed to work Ok without the patch. The right thing to do is to obviously read the specs but those are very long and I'm yet to find a good concise doc that explains proper init that works for older and newer cards.
Sorry.

JoNil:
I just ran into this issue as well, i tried the path in the closed pr and that soled it for me as well. I'm running on an raspberry pi pico with an old 4GB sd card.

thejpster:
Do you have (weak) pull-ups on your SPI lines? I had terrible trouble without them.

derchr:
The PR fixed it for me too, thanks!

mdarrik:
So if this is helpful, I tested the code in the [current version of the linked PR](https://github.com/jbeaurivage/embedded-sdmmc-rs/commit/3be7da402f57a26df21f6b910367b0f0ee7a77a8), and it didn't seem to work for me out of the box. What I found was that if I flashed an example using an Arduino example, this lib would work for me suddenly until I removed the card or the controller lost power. 

However, if I wrote `CMD0` over SPI before running the init, the card started working. 

Looking into SDFat a bit more, it looks like [it doesn't wait for ready if the command being sent is CMD0 or CMD12](https://github.com/greiman/SdFat/blob/628effa1c277e6dc1fef461b4b755ad300735248/src/SdCard/SdSpiCard.cpp#L253). So it might be worth merging the 2 approaches tried in the branch's history together: flushing the card & sending CMD0 without waiting for not busy.

thejpster:
Great. Any luck when testing that branch?

[...]

PR #65, opened 2022-09-26
Add the option to skip waiting in the init sequence when issuing CMD0.

jbeaurivage:
Looking at #32 and #33, at least some cards get stuck in `TimeoutWaitNotBusy` when trying to initialize them. #32 provides a fix, however it seems to break other cards' init sequence.

I suggest adding an option in `AcquireOpts` to use this modified init sequence, with the default being to use the "regular" sequence.

### DIFF
--- a/src/sdmmc.rs
+++ b/src/sdmmc.rs
@@ -129,11 +129,18 @@ impl Delay {
 pub struct AcquireOpts {
     /// Some cards don't support CRC mode. At least a 512MiB Transcend one.
     pub require_crc: bool,
+    /// Some cards should send CMD0 after an idle state to avoid being stuck in [`TimeoutWaitNotBusy`].
+    /// See [this conversation](https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/33)
+    /// and [this one])(https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/33).
+    pub skip_wait_not_busy: bool,
 }
 
 impl Default for AcquireOpts {
     fn default() -> Self {
-        AcquireOpts { require_crc: true }
+        AcquireOpts {
+            require_crc: true,
+            skip_wait_not_busy: false,
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/sdmmc.rs
```diff
@@ -128,11 +128,18 @@ impl Delay {
 pub struct AcquireOpts {
     /// Some cards don't support CRC mode. At least a 512MiB Transcend one.
     pub require_crc: bool,
+    /// Some cards should send CMD0 after an idle state to avoid being stuck in [`TimeoutWaitNotBusy`].
+    /// See [this conversation](https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/33)
```
thejpster, line 132: These two links are the same. Might be worth anchoring to a specific comment.

thejpster:
It would awkward if you needed two versions of your code to work with the widest range of SD cards. My general approach has been "Do what Arduino does" - so I checked, and they don't do this.

Their init code is: https://github.com/greiman/SdFat/blob/master/src/SdCard/SdSpiCard.cpp#L170

Maybe we should add this loop of sending 255 lots of `0xFF` to clear the card instead?

```cpp
  // command to go idle in SPI mode
  for (uint8_t i = 1;; i++) {
    if (cardCommand(CMD0, 0) == R1_IDLE_STATE) {
      break;
    }
    if (i == SD_CMD0_RETRY) {
      error(SD_CARD_ERROR_CMD0);
      goto fail;
    }
    // Force any active transfer to end for an already initialized card.
    for (uint8_t j = 0; j < 0XFF; j++) {
      spiSend(0XFF);
    }
  }
```

jbeaurivage:
That would be even better. Just to clarify, that sequence would always run, without an option to skip it? I'll try to test it out with a few cards next week.

thejpster:
Well, it doesn't happen if your card returns IDLE on the first attempt. But every time it doesn't return IDLE, it gets flushed.

thejpster:
Does this work? I guess `skip_wait_not_busy` is no longer required if so?

jbeaurivage:
I'm not sure if it works yet. Getting my cards to fail to initialize is somewhat hard; it happens intermittently and is difficult to reproduce reliably. I'll have to do some more testing, but what I can say is I haven't had an initialization fail using this new method. But you're right that `skip_wait_not_busy` won't be needed anymore if it works.

PR #67, opened 2022-10-17
Fix closing `File`s and `Directories`

### LINKED ISSUE #66: File `starting_cluster` has a value of zero within the `open_files`
I've found an issue when trying to close files and open them again. Basically what ends up happening is that upon opening the file, the [`starting_cluster` of the File initialized to zero](https://github.com/rust-embedded-community/embedded-sdmmc-rs/blob/develop/src/controller.rs#L365), and then this zero value [is subsequently set on the `open_files`](https://github.com/rust-embedded-community/embedded-sdmmc-rs/blob/develop/src/controller.rs#L373).

At some point the `starting_cluster` value of the File is adjusted, but the value stored within `open_files` is not. 

Which in turn causes the [`close_file` function within the controller](https://github.com/rust-embedded-community/embedded-sdmmc-rs/blob/develop/src/controller.rs#L576) to not close the file.

If you attempt to open/close files multiple times (beyond the specified `MAX_FILES`) value, the library returns an Error.

I'm trying to figure out a solution to this at the moment. Any help is appreciated. Thanks

jbeaurivage:
This PR attempts to fix closing `Files` and `Directories`.

The working principle is that each new `File` and `Directory` is assigned a unique `SearchId`. To close/delete them, instead of looking for magic `Cluster` numbers, we simply need to lookup the ID in the list of open files or directories. This mitigates the issues where we might forget to update the cluster number, or where there might be multiple cluster numbers with the same value (most notably, 0).

Also replace raw arrays with `heapless::Vec`s for a cleaner implementation.

Optionally, this could remove the need for a `&Volume` parameter in the `close_dir` and `close_file` methods.

@ostenning and @thejpster, I would very much like to know what you think of this. I'm not super familiar with the internals of this library so I'd definitely like a good review.

Hopefully fixes #66.

### DIFF
--- a/src/filesystem/search_id.rs
+++ b/src/filesystem/search_id.rs
@@ -1,3 +1,5 @@
+use core::num::Wrapping;
+
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
 /// Unique ID used to search for files and directories in the open File/Directory lists
@@ -7,19 +9,21 @@ pub struct SearchId(pub(crate) u32);
 ///
 /// This object will always return a different ID.
 pub struct IdGenerator {
-    next_id: u32,
+    next_id: Wrapping<u32>,
 }
 
 impl IdGenerator {
     /// Create a new [`IdGenerator`].
     pub const fn new() -> Self {
-        Self { next_id: 0 }
+        Self {
+            next_id: Wrapping(0),
+        }
     }
 
     /// Generate a new, unique [`SearchId`].
     pub fn get(&mut self) -> SearchId {
         let id = self.next_id;
-        self.next_id = self.next_id.wrapping_add(1);
-        SearchId(id)
+        self.next_id += 1;
+        SearchId(id.0)
     }
 }
--- a/src/filesystem/search_id.rs
+++ b/src/filesystem/search_id.rs
@@ -1,0 +1,34 @@
+#[derive(Clone, Copy, Debug)]
+#[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
+/// Unique ID used to search for files and directories in the open File/Directory lists
+pub struct SearchId(pub(crate) u32);
+
+impl PartialEq for SearchId {
+    fn eq(&self, other: &Self) -> bool {
+        self.0 == other.0
+    }
+}
+
+/// ID generator intented to be used in a static context.
+///
+/// This object will always return a different ID.
+pub struct IdGenerator {
+    next_id: core::sync::atomic::AtomicU32,
+}
+
+impl IdGenerator {
+    /// Create a new [`IdGenerator`].
+    pub const fn new() -> Self {
+        Self {
+            next_id: core::sync::atomic::AtomicU32::new(0),
+        }
+    }
+
+    /// Generate a new, unique [`SearchId`].
+    pub fn next(&self) -> SearchId {
+        use core::sync::atomic::Ordering;
+        let id = self.next_id.load(Ordering::Acquire);
+        self.next_id.store(id + 1, Ordering::Release);
+        SearchId(id)
+    }
+}
--- a/src/filesystem/search_id.rs
+++ b/src/filesystem/search_id.rs
@@ -1,13 +1,7 @@
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
 /// Unique ID used to search for files and directories in the open File/Directory lists
 pub struct SearchId(pub(crate) u32);
-
-impl PartialEq for SearchId {
-    fn eq(&self, other: &Self) -> bool {
-        self.0 == other.0
-    }
-}
 
 /// ID generator intented to be used in a static context.
 ///
--- a/src/filesystem/search_id.rs
+++ b/src/filesystem/search_id.rs
@@ -7,22 +7,19 @@ pub struct SearchId(pub(crate) u32);
 ///
 /// This object will always return a different ID.
 pub struct IdGenerator {
-    next_id: core::sync::atomic::AtomicU32,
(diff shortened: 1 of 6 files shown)


thejpster:
As far as I recall, I picked cluster numbers because every directory and every file must be located within a unique cluster number, and there are no duplicates across any other file or directory on that Volume.

thejpster:
Hmm. As noted in the issue, I can see a problem where the volume is correct and multiple directory entries point to a file with the same starting cluster.

src/filesystem/search_id.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub const fn new() -> Self {
+        Self {
+            next_id: core::sync::atomic::AtomicU32::new(0),
+        }
+    }
+
+    /// Generate a new, unique [`SearchId`].
+    pub fn next(&self) -> SearchId {
+        use core::sync::atomic::Ordering;
+        let id = self.next_id.load(Ordering::Acquire);
```
thejpster, line 24: This isn't thread safe. You either need to use a critical-section (from the critical-section crate) or do fetch-add (which doesn't work on Cortex-M0 or RV32IMC).

thejpster: Also, unless you can prove otherwise, it's safest to use SeqCst ordering always. The consequences of getting it wrong are tricky subtle bugs, and the benefits of getting it right are maybe a slight performance bump that is irrelevant in a cold path of a microcontroller based filesystem.

thejpster (changes requested):
Heapless Vec makes much more sense here. But there's an issue with the ID generation.

src/filesystem/search_id.rs
ostenning: For thread safety, wouldn't it be more appropriate to have the ID generator as a member of the controller, rather than a static global? I believe the open files and (therefore the unique search ID) are specific to the controller only.

jbeaurivage: You're right. I made `Controller` own an `IdGenerator` instead of a global static generator.

jbeaurivage:
> As far as I recall, I picked cluster numbers because every directory and every file must be located within a unique cluster number, and there are no duplicates across any other file or directory on that Volume.

In that case, seems like sticking with `Vec` makes sense, but keeping the ID generator shouldn't be necessary. However as mentioned in #66, files are initially created with a cluster number of 0, which makes them not searchable within the `Vec` until they have a "real" cluster number assigned. I added code to update the cluster number in the open files list when one gets assigned, but that doesn't fully fix the problem since there's a transient state where the file still is assigned to cluster 0.

src/filesystem/search_id.rs
```diff
@@ -0,0 +1,25 @@
...
+impl IdGenerator {
+    /// Create a new [`IdGenerator`].
+    pub const fn new() -> Self {
+        Self { next_id: 0 }
+    }
+
+    /// Generate a new, unique [`SearchId`].
+    pub fn get(&mut self) -> SearchId {
+        let id = self.next_id;
+        self.next_id += 1;
```
thejpster, line 22: In debug mode this will crash after 4 billion file open operations. In release mode it will silently wrap. As wrapping is probably fine, we should explicitly do a 'wrapping add'.

thejpster:
> Unfortunately this is where my knowledge of FAT is too limited to move forward; can't files be assigned a real cluster number as soon as they're created, and not when first written to?

Yes it's probably fine to allocate the first cluster when you create a new file.

elpiel:
@jbeaurivage is there any progress on this PR?
I've found some other issues related to opening/closing files which you can check here: #75

jbeaurivage:
@elpiel unfortunately I haven't had the time to jump back into that issue. Thanks for reporting #75, I'll try to look at that when I finally manage to work on this.

jbeaurivage:
All that is left to do at this point is trying to do away with the IdGenerator and try to have unique clusters for every file and directory that is created, instead of going through the zero-cluster first.

src/filesystem/search_id.rs
```diff
@@ -0,0 +1,25 @@
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
+/// Unique ID used to search for files and directories in the open File/Directory lists
+pub struct SearchId(pub(crate) u32);
+
+/// ID generator intented to be used in a static context.
+///
+/// This object will always return a different ID.
+pub struct IdGenerator {
+    next_id: u32,
```
thejpster, line 10: I know this is going away, but if this was a WrappingU32, you wouldn't need to remember to call .wrapping_add on it.

src/volume_mgr.rs
```diff
@@ -666,6 +737,19 @@ where
     }
 }
 
+fn cluster_position_by_id(list: &[ClusterDescriptor], id_to_find: SearchId) -> Option<usize> {
+    list.iter().position(|f| f.compare_id(id_to_find))
+}
+
+fn cluster_already_open(
+    vec: &[ClusterDescriptor],
```
thejpster, line 745: Maybe call this `cluster_list`, not vec?

src/volume_mgr.rs
```diff
@@ -666,6 +737,19 @@ where
     }
 }
 
+fn cluster_position_by_id(list: &[ClusterDescriptor], id_to_find: SearchId) -> Option<usize> {
```
thejpster, line 740: Maybe call this the same as whatever you call it in the function below?

thejpster:
Looking good! If we don't have any tests for opening multiple files, or opening the same file twice (the second open should fail), we should add some.

thejpster:
If we can resolve the two nits, I'd be inclined to merge this and maybe look at making IdGenerator go away later.

jbeaurivage:
Sounds good! Let me try to resolve the last suggestions today. By the way, there is a test that attempts to open the same file twice. Up until the last commit that test was failing (file could be opened twice without errors), but it passes now, so we're covered there :)

thejpster (approved):

PR #73, opened 2023-02-06
Skip wait init sequence

elpiel:
As discussed in issue #33, I've tested this with a brand new card from SanDisk (64GB) and so far it's working.

As mentioned in comment https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/33#issuecomment-1407425962, there are a few responses that are returned first (probably responses for proprietary functionality) and after the initial 2 responses, it acquires the device successfully.

```
0.001253 WARN  Got response: 3f, trying again..
└─ embedded_sdmmc::sdmmc::{impl#2}::acquire_with_opts::{closure#0} @ /mnt/Disk/Projects/lechev.space/embedded-sdmmc-rs/src/sdmmc.rs:211
0.001429 WARN  Got response: 7f, trying again..
└─ embedded_sdmmc::sdmmc::{impl#2}::acquire_with_opts::{closure#0} @ /mnt/Disk/Projects/lechev.space/embedded-sdmmc-rs/src/sdmmc.rs:211
```

### DIFF
--- a/src/sdmmc.rs
+++ b/src/sdmmc.rs
@@ -129,11 +129,18 @@ impl Delay {
 pub struct AcquireOpts {
     /// Some cards don't support CRC mode. At least a 512MiB Transcend one.
     pub require_crc: bool,
+    /// Some cards should send CMD0 after an idle state to avoid being stuck in [`TimeoutWaitNotBusy`].
+    /// See [this conversation](https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/33)
+    /// and [this one])(https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/33).
+    pub skip_wait_not_busy: bool,
 }
 
 impl Default for AcquireOpts {
     fn default() -> Self {
-        AcquireOpts { require_crc: true }
+        AcquireOpts {
+            require_crc: true,
+            skip_wait_not_busy: false,
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


elpiel:
@thejpster I've been using this code for some time now but all of the sudden I started to get these issues with acquiring and writing to the sd card.

Like CRC errors `CrcError(39826, 33169)`, `CantEnableCRC` to Timeout errors for CMD0 and CMD8.

I'm not sure how I can check what's going wrong and I've already tried with different devices (i.e. a different pico and the same board with SD card but a new one) all resulting in the same issue.

thejpster:
What SPI clock speed are you running at? Does going slower make it better?

elpiel:
I was running at 33 MHz and tried at 30 MHz and 15 MHz but the result is the same.

elpiel:
@thejpster  Is it possible that running and stopping the board while having opened files and dirs might affect the next time I run an application on the device?

I create a logger for the SD card and open 2 files which I keep the struct for `&'static` with `StaticCell`. Right now I started to get `DirAlreadyOpen` error even though there is a single dir opened in the while application and that's `root`

thejpster:
I haven't tested writing to files much - my use case was read only. There are almost certainly bugs, and bugs can lead to filesystem errors that need Scandisk/fsck to fix, or a reformat.

elpiel:
Let's continue with this PR, it seems to be working @thejpster .
I will make issue for the other problems I've encountered.

elpiel:
> FAT is a non-journalled file system so you really should be closing files as soon as possible, and ideally never reboot with files open. Although I think this library is designed to not cache too much stuff, I haven't checked it in detail.
Does this statement also applies for Directories? I.e. can I open the root dir for the whole life of the application without consequences to the filesystem?

thejpster:
Opening things without writing to them should be fine. I believe it's updating the directory entry (or not) for a file when you close the file that's the issue.

elpiel:
@thejpster would you consider merging this PR?

elpiel:
Since this branch is outdated and I have no access to @mdarrik repo, I will open a new PR.

PR #38, opened 2021-08-13
Allow initializing without CRC

dcz-self:
A 512MB Transcend card seems to not support it but work OK otherwise:
MID: 0x00001b
OID: 0x534d
PNM: SMI
MDT: 11/2007

PR #80, opened 2023-05-06
Big API change

thejpster:
Now the public API only has `SdCard`. This object uses `&self` references and does auto-initialisation of the card.

Internally it uses a `RefCell` and defers to some inner type which takes `&mut self` references.


thejpster:
image

Seems to read OK.

elpiel:
I'll jump in the PR to add some improvements I've been doing locally.
It would be good to put up an `-async` wip crate in the repo in a subsequent PR. I think it's time to introduce an upstream version in this repo.
E.g. https://docs.embassy.dev/embassy-rp/git/rp2040/spi/struct.Spi.html#implementations

For embassy, some members of AeroRust shared that they are working on their own virtual filesystem and I use littlefs2 so API should be modular to accommodate additional filesystems:

https://github.com/McMaster-Rocketry-Team/rust-monorepo/tree/master/vlfs

Enum doesn't seem like a good choice for this and for example, we can use the type system instead.

thejpster:
Ok but let's not do too much in one go. If this is good, let's get it in and look at async and the like separately.

thejpster:
I also want to make clear - it is not my intention for this crate to "be an Operating System". As I see it, the job of an Operating System is to manage multiple disks with multiple filing systems and provide some kind of abstraction over that. This crate may be useful to anyone who wants to write an Operating System, but it is just a crate that speaks SD/MMC protocol over an SPI bus, and which can read/write files on an FAT32 formatted SD card.

If someone wants to take the block device traits and put them in another crate, then replace the code here with a link to that crate, that would be amazing. If someone wants to implement a Virtual Filing System trait in another crate, happy to look at importing and implementing it. But this crate isn't the place for that.

elpiel:
Yes, I totally agree on that.
I suppose a good candidate for that is `embedded-storage`

https://github.com/rust-embedded-community/embedded-storage
There's also a crate for the `async` API interface.

I haven't found a good approach to introduce both interfaces (yet) - blocking and async in this current version of the SdCard.

thejpster:
Hmm, yes and no.

| Kind | Min Read Size | Min Write Size | Min Erase Size |
|------|---------------|----------------|----------------|
| NOR Flash | 1 byte | 1 byte | 4K Erase Page |
| NAND Flash | 128 byte | 128 byte | 4K Erase Page |
| I2C EEPROM | 1 byte | 1 byte | N/A |
| SD Card | 512 byte | 512 byte | N/A |

These are typical values - NAND sector sizes can vary (sectors also have extra bytes for storing ECC) and Erase Page sizes can vary too.

src/sdcard/mod.rs
```diff
@@ -5,176 +5,387 @@
...
+        }
+        self.spi.transfer(buffer).map_err(|_e| Error::Transport)?;
+
+        let mut crc = u16::from(self.receive()?);
+        crc <<= 8;
+        crc |= u16::from(self.receive()?);
+
+        let calc_crc = crc16(buffer);
+        if crc != calc_crc {
+            return Err(Error::CrcError(crc, calc_crc));
```
jannic, line 341: Shouldn't crc checking be optional? Or what's the use case of the `require_crc` flag?

src/sdcard/mod.rs
```diff
@@ -206,89 +417,73 @@ where
                     Ok(R1_IDLE_STATE) => {
                         break;
                     }
-                    Ok(r) => {
+                    Ok(_r) => {
                         // Try again
-                        warn!("Got response: {:x}, trying again..", r);
+                        warn!("Got response: {:x}, trying again..", _r);
+                        attempts -= 1;
```
jannic, line 423: `delay` already contains a retry count. With this change, that becomes entirely redundant.
The retry count of `delay` is much higher though, 32000.

thejpster:
These are all excellent observations. I'll attend to them in my OSS time on Friday.

src/sdcard/mod.rs
thejpster: I saw this and did wonder what was going on. Making CRC optional was added in #38.

src/sdcard/mod.rs
thejpster: Fair point, well made. I've changed it to use the Delay max retry count, and set that to a more reasonable max 512 retries.

src/sdcard/mod.rs
thejpster: This is now changed to be `use_crc` and if it's false, we don't enable the CRC and we don't do any CRC calculations, sending or receiving. If it's true we enable the CRC with CMD59 and we do send/receive the CRC16s on all blocks.

src/sdcard/mod.rs
```diff
@@ -338,274 +533,134 @@ where
...
+    SDHC,
+}
 
-        let mut crc = u16::from(self.0.receive()?);
-        crc <<= 8;
-        crc |= u16::from(self.0.receive()?);
+/// A terrible hack for busy-waiting the CPU while we wait for the card to
+/// sort itself out.
+///
+/// @TODO replace this!
```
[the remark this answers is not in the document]
thejpster: I cleaned up all the comments on this type.

thejpster:
OK, I attended to them today instead of waiting until Friday. Hopefully it's cleaner, faster and better documented!

thejpster:
Aaaand now the tests should pass as well.

README.md
```diff
@@ -12,14 +12,10 @@ You will need something that implements the `BlockDevice` trait, which can read
 
 ```rust
 // Build an SD Card interface out of an SPI device
-let mut spi_dev = embedded_sdmmc::SdMmcSpi::new(sdmmc_spi, sdmmc_cs);
-// Try and initialise the SD card
-let block_dev = spi_dev.acquire()?;
-// The SD Card was initialised, and we have a `BlockSpi` object
-// representing the initialised card.
-write!(uart, "Card size is {} bytes", block_dev.card_size_bytes()?)?;
+let sdcard = embedded_sdmmc::SdCard::new(sdmmc_spi, sdmmc_cs);
```
jannic, line 15: The delayer parameter is missing in this example.

jannic: You can include README in the doc tests by adding this to lib.rs:
```
// Include README.md so its doc tests are run
#[doc = include_str!("../README.md")]
#[cfg(doctest)]
pub struct ReadmeDoctests;
```
But that triggers quite a lot of errors, because the examples in the README use some variables without initializing them. Of course, one could add those. But that would be a lot of boilerplate. In normal doc comments, that's not a problem, because this initialization code can be hidden from the documentation. That's not possible in the README.

thejpster: Copied the README example into an example file and made sure it actually compiles.

thejpster:
Is this now good to go @jannic?

jannic:
I'm probably not the right person to ask. My only use case for this crate is the neotron, and you know that one much better than I do. Porting the example from rp2040-hal to new API was easy, but that also doesn't add much real world experience. So some validation on a different architecture or at least different microcontroller would be nice.

jonathanpallant:
@jannic but did I at least address all of your comments?

src/sdcard/mod.rs
jannic: 512 retries with a 10µs delay each are ~5ms. I guess that's enough in most cases, but I'm not sure if it's sufficient in general: For example, the same retry count is used when waiting for the response to a write command. What if a write command needs to perform some defragmentation due to the card's write leveling algorithm?
I didn't check the spec to see if there are maximum delays specified. But the old value of 32000 retries / 0.3s, while rather long, isn't necessarily wrong. Without knowing why the value was chosen that high in the first place, I'd be worried that it might be needed in some cases. (Chesterton's fence...)

So it might be better to keep the old value as a default, and use `Delay::new( lower_value )` in cases where a shorter timeout is desired?

src/sdcard/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+    /// Write an arbitrary number of bytes to the card using the SD protocol and
+    /// an optional CRC.
+    fn write_data(&mut self, token: u8, buffer: &[u8]) -> Result<(), Error> {
+        self.write_byte(token)?;
+        self.write_bytes(buffer)?;
+        if self.options.use_crc {
+            let calc_crc = crc16(buffer);
+            let bytes = calc_crc.to_be_bytes();
+            self.write_bytes(&bytes)?;
+        }
```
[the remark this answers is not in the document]
thejpster: Ah. This does work though - I wonder if they get absorbed by a subsequent polling loop. I'll fix it though.

9names (dismissed):
I have nothing to add. New comments are good, new structure looks cleaner to me.

src/sdcard/mod.rs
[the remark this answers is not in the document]
thejpster: Fixed

thejpster:
I did a rebase, so it cleared your review @9names. Sorry!

jannic (approved):

thejpster:
Sorry @jannic, it seems you don't have write access here.

9names (approved):
Re-approving, I don't think have write approval here either so not sure it matters.

thejpster:
Good enough for me. Admin override.

PR #83, opened 2023-05-14
Add file extension filter.

YSawc:
Add file extension filtering for directory iterator function. There may be need to refactor code in my adding iteration with file extension filter function because of almost the same as original iterator function, but I think it is next phase of comment and I'm just little out of idea.

### DIFF
--- a/src/fat/ondiskdirentry.rs
+++ b/src/fat/ondiskdirentry.rs
@@ -124,8 +124,13 @@ impl<'a> OnDiskDirEntry<'a> {
     }
 
     /// Does this on-disk entry match the given filename?
-    pub fn matches(&self, sfn: &ShortFileName) -> bool {
+    pub fn file_name_matches(&self, sfn: &ShortFileName) -> bool {
         self.data[0..11] == sfn.contents
+    }
+
+    /// Does this on-disk entry match the given extention?
+    pub fn extension_matches(&self, extension: &str) -> bool {
+        &self.data[8..11] == extension.as_bytes()
     }
 
     /// Which cluster, if any, does this file start at? Assumes this is from a FAT32 volume.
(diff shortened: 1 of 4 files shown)


src/fat/ondiskdirentry.rs
```diff
@@ -124,10 +124,15 @@ impl<'a> OnDiskDirEntry<'a> {
     }
 
     /// Does this on-disk entry match the given filename?
-    pub fn matches(&self, sfn: &ShortFileName) -> bool {
+    pub fn file_name_matches(&self, sfn: &ShortFileName) -> bool {
         self.data[0..11] == sfn.contents
     }
 
+    /// Does this on-disk entry match the given extention?
```
thejpster, line 131:
```suggestion
    /// Does this on-disk entry match the given extension?
```

thejpster:
First of all, many thanks for sending in a PR! We really appreciate it when people take the time to contribute.

On first glance, it looks like code you added is a copy-paste of the existing `iterate_dir` function, but with the extra code to filter for file extensions. I think we'd need to see this written in a way that doesn't duplicate quite so much code, to keep it maintainable and avoid duplicating bugs.

I'm also not clear if we really need this function, as I think you could write something like:

```rust
volume_mgr
    .iterate_dir(&volume, &root_dir, |x| {
        if x.name.extension != b"TXT" { return; }
        println!("\t\tFound: {:?}", x);
    })?;
```

(I haven't tested this though).

Do you have an example of how you would want to use this new function?

YSawc:
Thanks feedback! I overlooked existing `extension()` function..

I found below code works filtering of file extension! About of described code, `extension` field of name struct not existing but `extension()` function exist.
```
            volume_mgr
                .iterate_dir(&volume, &root_dir, |x| {
                    if x.name.extension() != b"NES" {
                        return;
                    }
                    println!("\t\tFound: {:?}", x);
                })
                .unwrap();
```
Now this PR no more no need so I close this!

PR #85, opened 2023-05-20
Removed unused field.

thejpster:
It was a rustc warning and it seems to work without. If we need it in future, we can add it back.


thejpster:
Might as well fix this, because it's annoying.

thejpster:
Pinging @rust-embedded-community/all

eldruin (approved):
Nice, thanks!

PR #89, opened 2023-06-11
User-selectable number of retries for card acquisition

### LINKED ISSUE #88: Make number of card init retries user selectable.
Following up on the conversation in #64. It would be nice to have a way for the user to specify the number of times the library should retry initializing the card before giving up and returning an error.

jbeaurivage:
Provides a way for the user to specify how many times acquiring the card should be retried before giving up and returning `Error::CardNotFound`. This is done by way of a field in `AcquireOpts`.

Closes #88.

### DIFF
--- a/src/sdcard/mod.rs
+++ b/src/sdcard/mod.rs
@@ -715,13 +715,12 @@ impl Delay {
         T: embedded_hal::blocking::delay::DelayUs<u8>,
     {
         if self.retries_left == 0 {
-            return Err(err);
+            Err(err)
         } else {
             delayer.delay_us(10);
             self.retries_left -= 1;
+            Ok(())
         }
-
-        Ok(())
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/sdcard/mod.rs
```diff
@@ -708,12 +715,13 @@ impl Delay {
         T: embedded_hal::blocking::delay::DelayUs<u8>,
     {
         if self.retries_left == 0 {
-            Err(err)
+            return Err(err);
         } else {
```
thejpster, line 719: If the first block has a `return` in it, you don't need the else.

thejpster:
One observation, but definitely I agree in principle.

src/sdcard/mod.rs
jbeaurivage: Oh, yeah, that's a misrefactor. Let me fix it up quickly.

thejpster (approved):

PR #92, opened 2023-08-26
Add a local on-stack FAT cache for next_cluster

ftilde:
For some operations, e.g. seeking in a file, this makes a huge performance difference since we avoid (re)reading the FAT for every node in the linked list.
For other operations that call next_cluster only once there are no negative implications since the memory for the FAT block would have been allocated on the stack anyways.

A motivation from my specific use case: Seeking in a somewhat large audio file (e.g. 30 minutes) would take multiple seconds, but is now without any perceptible delay.

### DIFF
--- a/src/fat/mod.rs
+++ b/src/fat/mod.rs
@@ -14,6 +14,38 @@ pub enum FatType {
     Fat32,
 }
 
+pub(crate) struct BlockCache {
+    block: Block,
+    idx: Option<BlockIdx>,
+}
+impl BlockCache {
+    pub fn empty() -> Self {
+        BlockCache {
+            block: Block::new(),
+            idx: None,
+        }
+    }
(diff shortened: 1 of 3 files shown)


thejpster:
Interesting!

I'll check this out on Friday when I have some free time.

thejpster:
Still on my todo list. I got distracted by other things.

thejpster:
OK, picking this up right now!

jonathanpallant (approved):
I measure a small increase in stack usage (about 96 bytes) but that seems OK.

jonathanpallant:
Thank you for this PR! I'll schedule a release soon once I've had another pass to check if there's anything else we should be caching.

ftilde:
Excellent, thank you!

Issue #86, opened 2023-05-21
Use doc(inline)


thejpster:
https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#inline-and-no_inline

If you `pub use` a type adding `#[doc(inline)]` documents it like it lives where the use is, instead of listing it as an import.

Issue #75, opened 2023-02-24
Multiple opened files with empty content triggers `DirAlreadyOpen`


elpiel:
Trying to open multiple empty files at the same time with `Controller::open_file_in_dir` in Mode different than `ReadWriteCreate` triggers `open_dir_entry` causes a `DirAlreadyOpen` error.

Is returning `DirAlreadyOpen` a correct response here as well? It seems that it's going over opened files and yet returns `DirAlreadyOpen`?

thejpster:
Damn, I think this is still a bug.

thejpster:
OK fixed it.

eldruin:
Can this be closed now?

elpiel:
I believe so. I saw the issue mentioned in the last PR.
Closing this issue, if any new problems arise someone can open a new issue.

Issue #94, opened 2023-09-15
Writing a file doesn't update directory entry


jonathanpallant:
We need to go back and update the directory entry when the file is closed. Otherwise you can write bytes to a file but the directory entry says you didn't.

jonathanpallant:
At the same time we might as well defer writing to the FAT32 info sector.

Issue #90, opened 2023-06-21
get_card_type() should trigger card acquisition


jonathanpallant:
We forgot to try and initialise the card when that method is called.

PR #95, opened 2023-09-15
Cleaning up some comments, and fixing the close file bug.

### LINKED ISSUE #87: Acquire SD card
I've come to realise that in version `0.5` in order to acquire the SD card you need to call `check_init` or `acquire` but it's not exposed from `SdCardInner`.
Instead you need to call either `num_bytes` or directly try to read/write to the SD card in order to initialize it.
It would be nice to be able to expose `acquire` or better the `check_init` function in `SdCard` in order to be able to initialize it upfront before the first read/write takes place.

jonathanpallant:
Whilst I was tidying up, I found #94. Let's fix it!

Also fixes #87 and #90


jonathanpallant:
Let's do #72 too!

jonathanpallant:
Fixes #86 too

jannic (approved):
I'm not a reviewer so it doesn't help much, but I read the patch and didn't find anything dubious. And many of the changes are obvious improvements.

Regarding the delayed metadata update: What are the implications if the file is not properly closed (either due to a bug or do to a crash / power failure before the file is closed)?
As FAT isn't known to be very crash-tolerant, this probably isn't a huge issue, but maybe the expected result should be documented? (eg. will the freshly written data be lost? Or can a fsck repair the filesystem without data loss?)

thejpster:
A failure to update the directory entry will mean that it looks like the file has X bytes when it in fact has Y bytes in it. There data is still there but if X < Y, then you won't be able to access it, or it X > Y then you will read garbage on the end of your file (and possibly get a read error if you fall off the end of the cluster chain). The clusters will have been allocated at the point you did the write, so at least nothing else will be able to write over your data. Deleting and re-creating the file will fix the issue so no major harm done.

I believe the same issue occurs with MS-DOS.

We could in future add a guard to the File object so that if you drop it without closing it you get a panic. That might be better than silently truncating the file.

thejpster:
I was adding a note about `Drop` support not being implemented on `File`. Then I noticed that it would basically be impossible because each `File` would need to borrow both the `VolumeManager` and the `Volume`.

So, five hours later, `File`, `Directory` and the new `Volume` are just integers uniquely identifying an object stored inside the Volume Manager. So if you want to implement `Drop` on `File` you now only have to keep a reference to `&RefCell<VolumeManager>` in your `DroppableFile` structure. Also you no longer need to keep passing that `&Volume` value, which is nicer.

src/lib.rs
```diff
@@ -101,21 +101,32 @@ pub mod fat;
...
 
+use filesystem::SearchId;
+
+#[doc(inline)]
 pub use crate::blockdevice::{Block, BlockCount, BlockDevice, BlockIdx};
+
+#[doc(inline)]
 pub use crate::fat::FatVolume;
+
+#[doc(inline)]
```
elpiel, line 112: Very good addition!

src/sdcard/mod.rs
```diff
@@ -49,12 +49,18 @@ where
 {
     /// Create a new SD/MMC Card driver using a raw SPI interface.
     ///
+    /// The card will not be initialised at this time. Initialisation is
```
elpiel, line 52: Another great addition to the docs!

elpiel (commented):
I've reviewed everything apart from the VolumeManager and everything looks good so far.
I'll leave the fat16/32 impls review to someone that has a bit more knowledge around it and actively uses the filesystem impl.

thejpster:
> I'll leave the fat16/32 impls review 

Hopefully that bit didn't change much.

jonathanpallant:
To allay (my, and other's) concerns I'm writing some more test cases and trying to get the tarpaulin coverage number up.

jonathanpallant:
OK, I should stop fiddling and just merge it now. Last call!

thejpster:
Closes #75

elpiel:
thank you @thejpster for spending time to work on the crate!
Amazing work!

src/volume_mgr.rs
```diff
@@ -426,176 +500,187 @@ where
...
                 // Remember this open file
-                self.open_files
-                    .push(ClusterDescriptor::new(
-                        volume.idx,
-                        file.starting_cluster,
-                        search_id,
-                    ))
-                    .map_err(|_| Error::TooManyOpenFiles)?;
+                unsafe {
+                    self.open_files.push_unchecked(file);
```
ryan-summers, line 529: Maybe make a note that we checked above, so the unsafe is fine.

src/volume_mgr.rs
```diff
@@ -644,32 +733,54 @@ where
...
+        }
+
+        let (file_info, file_idx) = found_idx.ok_or(Error::BadHandle)?;
+
+        if file_info.dirty {
+            let volume_idx = self.get_volume_by_id(file_info.volume_id)?;
+            match self.open_volumes[volume_idx].volume_type {
+                VolumeType::Fat(ref mut fat) => {
+                    debug!("Updating FAT info sector");
+                    fat.update_info_sector(&self.block_device)?;
```
ryan-summers, line 774: I wonder if we should do this more often than just on file close to alleviate issues with crashing firmware. Is there a periodic task where things get flushed to disk?

src/volume_mgr.rs
```diff
@@ -680,36 +791,114 @@ where
 
     /// Consume self and return BlockDevice and TimeSource
     pub fn free(self) -> (D, T) {
-        (self.block_device, self.timesource)
+        (self.block_device, self.time_source)
+    }
+
+    /// Check if a file is at End Of File.
+    pub fn file_eof(&self, file: File) -> Result<bool, Error<D::Error>> {
```
ryan-summers, line 801: Honestly these feel like they should be members of `File` and accept the `Volume` as an arg?

ryan-summers (commented):
Some general comments/feedback. Overall things look good, just some minor recommendations and/or various thoughts on API design.

src/volume_mgr.rs
ryan-summers: Maybe just adding an API to do it and tell the user to periodically handle it. _or_ check during file IO if too much time has passed since the last flush and just update things to make sure they stay "sort of" in sync? Does this help prevent corruption at all?

src/filesystem/files.rs
```diff
@@ -1,23 +1,46 @@
...
+///
+/// Do NOT drop this object! It doesn't hold a reference to the Volume Manager
+/// it was created from and cannot update the directory entry if you drop it.
+/// Additionally, the VolumeManager will think you still have the file open if
+/// you just drop it, and it won't let you open the file again.
+///
+/// Instead you must pass it to [`crate::VolumeManager::close_file`] to close it
+/// cleanly.
+///
+/// If you want your files to close themselves on drop, create your own File
```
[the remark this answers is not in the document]
thejpster: That's the issue. I can't think of a way to have multiple files holding mut references to the volume manager, and I'm not sure I want to add interior mutability. It's easy enough to wrap these types if you need, and I know people who have.

But at least it's documented now :)

src/filesystem/files.rs
```diff
@@ -46,24 +69,24 @@ pub enum Mode {
     ReadWriteCreateOrAppend,
 }
 
-impl File {
+impl FileInfo {
     /// Are we at the end of the file?
     pub fn eof(&self) -> bool {
-        self.current_offset == self.length
+        self.current_offset == self.entry.size
```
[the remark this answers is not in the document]
thejpster: DirEntry is the directory entry - the piece of the containing directory that describes the file. The size field is the file size as printed by DIR. I realised the length was stored twice so I de-duped it.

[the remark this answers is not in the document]
ryan-summers: Gotcha, I think I realized this later in the review too, thanks for the explanation :)

src/volume_mgr.rs
```diff
@@ -117,6 +111,16 @@ where
         const PARTITION_INFO_LBA_START_INDEX: usize = 8;
         const PARTITION_INFO_NUM_BLOCKS_INDEX: usize = 12;
 
+        if self.open_volumes.is_full() {
+            return Err(Error::TooManyOpenVolumes);
+        }
+
+        for v in self.open_volumes.iter() {
+            if v.idx == volume_idx {
+                return Err(Error::VolumeAlreadyOpen);
```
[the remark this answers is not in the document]
thejpster: The volume represents the whole FAT file system. Free cluster counts might be corrupted if I let you open a volume twice.

Dropping a handle is bad. But it's no worse than dropping a FILE* in a C program.

src/volume_mgr.rs
```diff
@@ -218,151 +217,210 @@ where
...
         }
 
         // Check it's not already open
-        if cluster_already_open(&self.open_dirs, volume.idx, dir_entry.cluster) {
-            return Err(Error::DirAlreadyOpen);
+        for d in self.open_dirs.iter() {
+            if d.volume_id == self.open_volumes[volume_idx].volume_id
+                && d.cluster == dir_entry.cluster
+            {
+                return Err(Error::DirAlreadyOpen);
```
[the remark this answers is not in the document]
thejpster: I think the directory is the least bad thing to open twice but let's be consistent.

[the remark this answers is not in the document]
thejpster: If we returned a duplicate handle, they might close one and use the remaining one which wouldn't work. That would need ref-counting or something.

src/volume_mgr.rs
thejpster: A flush API could work. It's equivalent to a close and reopen without losing your seek position.

src/volume_mgr.rs
thejpster: Hmm. I could take `&mut VolumeManager` as an arg in a File method, true. Its important you give the right VolumeManager but we can sort of check that now.

src/fat/volume.rs
```diff
@@ -512,60 +512,63 @@ impl FatVolume {
...
                 };
                 let dir_size = match dir.cluster {
-                    Cluster::ROOT_DIR => BlockCount(
-                        ((u32::from(fat16_info.root_entries_count) * 32) + (Block::LEN as u32 - 1))
-                            / Block::LEN as u32,
-                    ),
+                    ClusterId::ROOT_DIR => {
+                        let len_bytes =
+                            u32::from(fat16_info.root_entries_count) * OnDiskDirEntry::LEN_U32;
+                        BlockCount::from_bytes(len_bytes)
```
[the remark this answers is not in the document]
thejpster: I'll leave a ticket for this but it's a pain because you need so many pieces of data.

src/filesystem/directory.rs
```diff
@@ -28,13 +29,35 @@ pub struct DirEntry {
 }
 
 /// Represents an open directory on disk.
+///
+/// Do NOT drop this object! It doesn't hold a reference to the Volume Manager
+/// it was created from and if you drop it, the VolumeManager will think you
+/// still have the directory open, and it won't let you open the directory
+/// again.
```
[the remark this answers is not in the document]
thejpster: Mutable references to the volume manager require interior mutability or only a single file open at once.

It's no worse than the C API with FILE* or int file handles.

src/volume_mgr.rs
thejpster: That's #98

src/volume_mgr.rs
thejpster: This is #99

src/sdcard/mod.rs
```diff
@@ -105,17 +120,30 @@ where
     }
 
     /// Get the card type.
+    ///
+    /// This will trigger card (re-)initialisation.
     pub fn get_card_type(&self) -> Option<CardType> {
-        let inner = self.inner.borrow();
+        let mut inner = self.inner.borrow_mut();
+        inner.check_init().ok()?;
```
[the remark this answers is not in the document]
thejpster: We *check* we are init, so that users don't have to manually perform an init step. If the card *is* initialised, this causes no work to happen.

src/volume_mgr.rs
thejpster: done

thejpster:
Thank you for the thoughtful comments! Much appreciated.

thejpster:
@ryan-summers  if you're happy would you mind pressing approve? It's set to auto merge.

ryan-summers (approved):
Looks good to me, thanks for taking my comments into consideration! :) Don't know if I have approval on this repo, but I'll approve anyways.

ryan-summers:
@thejpster if you add me with maintainer or write access, that should solve things :) or just force merge if you'd prefer

thejpster:
Bumped all members up to write permission. It was set to Read for some reason.

PR #100, opened 2023-09-23
Add a little shell example.

thejpster:
You can walk around a disk image and do directory listings. I had to make some more things 'Debug' so I could print the filesystem state. Also our handling of ".." was not right so I fixed that.

### DIFF
--- a/examples/shell.rs
+++ b/examples/shell.rs
@@ -10,85 +10,54 @@ use crate::linux::{Clock, LinuxBlockDevice};
 
 mod linux;
 
-struct State {
+struct VolumeState {
     directory: Directory,
     volume: Volume,
+    path: Vec<String>,
 }
 
-fn main() -> Result<(), Error<std::io::Error>> {
-    env_logger::init();
-    let mut args = std::env::args().skip(1);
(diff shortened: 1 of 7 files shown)


eldruin (commented):
Thanks!
Could you add an entry to the changelog and have a look at the CI?

thejpster:
@eldruin  added as requested. Should be ready now.

examples/shell.rs
```diff
@@ -0,0 +1,252 @@
...
+impl Context {
+    fn current_path(&self) -> Vec<String> {
+        let Some(s) = &self.volumes[self.current_volume] else {
+            return vec![];
+        };
+        s.path.clone()
+    }
+
+    fn process_line(&mut self, line: &str) -> Result<(), Error<std::io::Error>> {
+        if line == "help" {
```
ryan-summers, line 34: FYI, https://docs.rs/logos/latest/logos/ is really nice for things like this :)

ryan-summers (approved):

examples/shell.rs
thejpster: Interesting! I'll check it out.

PR #103, opened 2023-10-07
Smart Handles

thejpster:
`Volume`, `Directory` and `File` now hold `&mut VolumeManager`, which gives them methods and a `Drop` impl. Old functionality available with `RawVolume`, `RawDirectory` and `RawFile` because with the smart handles you can't open more than one file at once.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog] and this project adheres to [Semantic 
 
 ## [Unreleased]
 
-* None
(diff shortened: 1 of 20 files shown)


thejpster:
Requires #100

thejpster:
#100 is in.

We might want to do a release before we merge this? What do we think?

thejpster:
I released 0.6.0. So this can be 0.7.0.

yanorei32:
I'm looking forward to this feature release. 🚀

eldruin (commented):
Thank you! Could you add an entry to the changelog about this?

thejpster:
Ahh, nuts. Yes of course @eldruin .

thejpster:
Done.

eldruin (approved):
Thank you!

PR #106, opened 2023-10-22
Update embedded hal to 1.0.0

AnyTimeTraveler:
This is based on [thejpster](https://github.com/rust-embedded-community/embedded-sdmmc-rs/commits?author=thejpster)'s Pull Request #103.

The changes are suprisingly trivial.  
Except for one occasion, only where-clauses needed to be adjusted.

I need to work with an ESP32C6, which has already updated to this version in a feature gate.

### DIFF
--- a/examples/readme_test.rs
+++ b/examples/readme_test.rs
@@ -94,11 +94,14 @@ impl From<embedded_sdmmc::SdCardError> for Error {
 }
 
 fn main() -> Result<(), Error> {
+    // BEGIN Fake stuff that will be replaced with real peripherals
     let spi_bus = RefCell::new(FakeSpiBus());
     let delay = FakeDelayer();
     let sdmmc_spi = embedded_hal_bus::spi::RefCellDevice::new(&spi_bus, DummyCsPin, delay);
     let sdmmc_cs = FakeCs();
     let time_source = FakeTimesource();
+    // END Fake stuff that will be replaced with real peripherals
+
     // Build an SD Card interface out of an SPI device, a chip-select pin and the delay object
     let sdcard = embedded_sdmmc::SdCard::new(sdmmc_spi, sdmmc_cs, delay);
     // Get the card size (this also triggers card initialisation because it's not been done yet)
--- a/examples/readme_test.rs
+++ b/examples/readme_test.rs
@@ -3,26 +3,45 @@
 //! We add enough stuff to make it compile, but it won't run because our fake
 //! SPI doesn't do any replies.
 
-struct FakeSpi();
+use core::cell::RefCell;
 
-impl embedded_hal::spi::SpiDevice<u8> for FakeSpi {
+use embedded_sdmmc::sdcard::DummyCsPin;
+
+struct FakeSpiBus();
+
+impl embedded_hal::spi::ErrorType for FakeSpiBus {
     type Error = core::convert::Infallible;
-    fn transfer<'w>(&mut self, words: &'w mut [u8]) -> Result<&'w [u8], Self::Error> {
-        Ok(words)
-    }
 }
 
-impl embedded_hal::spi::SpiDevice<u8> for FakeSpi {
-    type Error = core::convert::Infallible;
-    fn write(&mut self, _words: &[u8]) -> Result<(), Self::Error> {
+impl embedded_hal::spi::SpiBus<u8> for FakeSpiBus {
+    fn read(&mut self, _: &mut [u8]) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    fn write(&mut self, _: &[u8]) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    fn transfer(&mut self, _: &mut [u8], _: &[u8]) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    fn transfer_in_place(&mut self, _: &mut [u8]) -> Result<(), Self::Error> {
+        Ok(())
+    }
+
+    fn flush(&mut self) -> Result<(), Self::Error> {
         Ok(())
     }
 }
 
 struct FakeCs();
 
-impl embedded_hal::digital::OutputPin for FakeCs {
+impl embedded_hal::digital::ErrorType for FakeCs {
     type Error = core::convert::Infallible;
+}
+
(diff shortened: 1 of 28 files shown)


thejpster:
@AnyTimeTraveler would you mind looking at the pipeline failures?

fu5ha:
According to [the docs of `SpiDevice`](https://docs.rs/embedded-hal/1.0.0-rc.1/embedded_hal/spi/trait.SpiDevice.html#tymethod.transaction), the `SpiDevice` impl itself is meant to take ownership of the CS pin, and that the implementations must first assert CS pin, then perform transaction, then deassert CS pin for every call to `transaction` (and all other methods just defer to that implementation). This is incompatible with the current signature of `SdCard` in that it also expects the CS pin separately, in order to meet the requirement to clock with CS deasserted in order to get the SDMMC device to actually let go of the Spi bus, as mentioned in the `SdCard` docs and SD/MMC physical layer spi spec.

There's a couple possible ways to solve this. The one that would mean the least required changes to this library and I think should work well would be to encourage the use of `DummyCsPin` in the creation of the `SpiDevice` impl that gets passed to creation of `SdCard` such that the real CS pin is owned by `SdCard` and can be managed manually separately, as we desire.

AnyTimeTraveler:
> @AnyTimeTraveler would you mind looking at the pipeline failures?

Yes, I will. I'm currently still planning on testing this with my ESP32C6 to make sure that the update didn't break anything.

AnyTimeTraveler:
@fu5ha I looked at complying with this new way of doing transactions and have found them incompatible with how the crate uses the SPI CS pin at the moment.
I agree with your approach.
I have not looked into the SD Card SPI protocol in enough depth to determine, wether the CS pin actualy needs to be held active for as long as the crate is holding it.
Maybe there is a way to handle this differently?
I will most likely move forward eith the current implementation and just use a Dummy CS Pin.

fu5ha:
I've opened https://github.com/AnyTimeTraveler/embedded-sdmmc-rs/pull/1 which makes a public `DummyCsPin` and documents its use but you could also just cherry-pick [4d3aa31](https://github.com/AnyTimeTraveler/embedded-sdmmc-rs/pull/1/commits/4d3aa31e4d40b057c86b7aaec802799de4ddbacf) into this branch.

AnyTimeTraveler:
Thank you very much! I will try to find time to work on this tonight.

thejpster:
Another solution is to require two SpiDevice objects. One which actually selects the SD card, and one which does not. You would just send out dummy bytes to the second one to get the SD card into SPI mode.

thejpster:
Talking on the matrix channel, I'm told the correct solution is to take exclusive ownership of the `SpiBus` whilst we want to clock out the init sequence. How that works in practice I'm not sure.

AnyTimeTraveler:
@thejpster I have a pretty good idea how to implement the proper solution.
It'll probably take a week or so. I don't have much free time right now.

What matrix channel were you talking about? That sounds useful to join.

thejpster:
https://matrix.to/#/#rust-embedded:matrix.org

thejpster:
Have you had a chance to look at this? I think we're expecting Embedded HAL 1.0 to drop on 2023-12-28, with Rust 1.75.

AnyTimeTraveler:
Sorry, I wish I had more time at the moment, but I had to put this on hold.
I expect to have more time at the start of February.
I do want to finish this, so I might squeeze it in before Christmas, but I can't promise anything.

luojia65:
Any further works on this pull request? embedded-hal 1.0.0 (no -rc.1!) is released. If you don't feel like having enough time working on this, I can take forward; our team will test it on RustSBI and RISC-V firmware ecosystem when moving on. @AnyTimeTraveler

jakezhu9:
@luojia65 and me have followed your work to update embedded-hal dependency to version 1.0.0. We sent a pull request to your repository: https://github.com/AnyTimeTraveler/embedded-sdmmc-rs/pull/2. Please have a look, hope this helps :)

AnyTimeTraveler:
> Any further works on this pull request?

Not from me, but I see your PR and will check it out tonight. 
If it turns out that more work is required, I would not mind at all @luojia65 taking over :)

AnyTimeTraveler:
Just tested locally and this should be ready to merge. 
All tests are passing :)

examples/readme_test.rs
```diff
@@ -74,9 +94,10 @@ impl From<embedded_sdmmc::SdCardError> for Error {
 }
 
 fn main() -> Result<(), Error> {
-    let sdmmc_spi = FakeSpi();
-    let sdmmc_cs = FakeCs();
+    let spi_bus = RefCell::new(FakeSpiBus());
```
jonathanpallant, line 97: Please change the README to match :)

src/lib.rs
```diff
@@ -22,47 +22,56 @@
...
-//! let mut root_dir = volume0.open_root_dir()?;
-//! let mut my_file = root_dir.open_file_in_dir("MY_FILE.TXT", embedded_sdmmc::Mode::ReadOnly)?;
-//! while !my_file.is_eof() {
-//!     let mut buffer = [0u8; 32];
-//!     let num_read = my_file.read(&mut buffer)?;
-//!     for b in &buffer[0..num_read] {
-//!         print!("{}", *b as char);
-//!     }
-//! }
+//! # let sdcard = embedded_sdmmc::SdCard::new(sdmmc_spi, sdmmc_cs, delayer);
```
jonathanpallant, line 61: This is the example code we want to see in the `cargo doc` output, so I don't think we should hide it with `#` marks?

examples/readme_test.rs
AnyTimeTraveler: The stuff you highlighted are the fake devices that will be replaced by real hardware in actual use cases.
The remainder of test is the same as in the Readme.md

src/lib.rs
AnyTimeTraveler: I have removed the #'s :)

AnyTimeTraveler:
This PR should be good to go now. 
Please poke me, if I have missed something :)

jonathanpallant (approved):

jonathanpallant:
Thank you!

Issue #107, opened 2023-11-08
No no-std example


Nereuxofficial:
There is no example for an embedded microcontroller(With the closest one having some FakeSPI structs). 

This is something i struggled with quite a bit during the making of [this repo](https://github.com/Nereuxofficial/esp-sdcard) which showcases using it on an ESP32. Most microcontrollers have quite similar approaches to doing this so i would suggest adding this(or something similar) to the repository examples to help people who want to use it on a no-std device.

I could also make a Pull Request for this but wanted to make sure this is wanted first.

jonathanpallant:
Correct, there are no microcontroller-specific examples here. I propose we don't include them because:

* they would have to go in a separate tree because we don't want the microcontroller-specific dependencies affecting the crate itself
* the HALs change and I don't have time to keep on top of them
* How do you pick which microcontrollers or boards to make a demo of?

In terms of who supports using this crate, https://crates.io/crates/embedded-sdmmc/reverse_dependencies gives some clues.

Looks like rp-rs have one, stm32-rs have one, and atsamd-rs have one. Happy to see a PR to link to their examples.

One example is https://github.com/rp-rs/rp-hal-boards/blob/main/boards/rp-pico/examples/pico_spi_sd_card.rs

PR #108, opened 2023-11-08
Linked no-std examples

Nereuxofficial:
As mentioned in #107 this would directly link some no-std examples of using the crate, making it easier for new users to get started


eldruin (approved):
Good idea, thank you!

PR #109, opened 2023-12-05
Fix struct reference in documentation

yanorei32:
I tried to work with the smart handles version of this library.
But I got a warning from `cargo doc`.

```
warning: unresolved link to `File`
  --> src/filesystem/directory.rs:54:47
   |
54 |     /// Convert a raw file into a droppable [`File`]
   |                                               ^^^^ no item named `File` in scope
   |
   = help: to escape `[` and `]` characters, add '\' before them like `\[` or `\]`
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: `embedded-sdmmc` (lib doc) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
```

This PR fixes it.

### DIFF
--- a/src/filesystem/directory.rs
+++ b/src/filesystem/directory.rs
@@ -51,7 +51,7 @@ pub struct DirEntry {
 pub struct RawDirectory(pub(crate) SearchId);
 
 impl RawDirectory {
-    /// Convert a raw file into a droppable [`File`]
+    /// Convert a raw file into a droppable [`crate::filesystem::File`]
     pub fn to_directory<
         D,
         T,


thejpster:
Damn. We should build the docs in the github actions.

Thank you for the fix!

src/filesystem/directory.rs
```diff
@@ -51,7 +51,7 @@ pub struct DirEntry {
 pub struct RawDirectory(pub(crate) SearchId);
 
 impl RawDirectory {
-    /// Convert a raw file into a droppable [`File`]
+    /// Convert a raw file into a droppable [`crate::filesystem::File`]
```
thejpster, line 54: I think it should be fixed like this instead:

```suggestion
    /// Convert a raw directory into a droppable [`Directory`]
```

thejpster (approved):

PR #110, opened 2023-12-17
Fix issue 74

### LINKED ISSUE #74: u32 overflow in finding fat_offset from cluster
Screenshot_20230216_230927_Chrome

I only have this message from a device deployed in the field. It seems to crash on:

```
                let fat_offset = cluster.0 * 4;

```

where everything involved is `u32`s.

thejpster:
Map ROOT_DIR (0xFFFF_FFFC) to an actual cluster number, so if you go off the end of the first cluster of the root directory on FAT32, the code no longer crashes trying to convert the ROOT_DIR magic cluster number into a disk offset.

Also adds an example which replicates the issue by writing an infinite number of files to the root directory.

Closes #74

### DIFF
--- a/examples/big_dir.rs
+++ b/examples/big_dir.rs
@@ -1,0 +1,39 @@
+extern crate embedded_sdmmc;
+
+mod linux;
+use linux::*;
+
+use embedded_sdmmc::{Error, VolumeManager};
+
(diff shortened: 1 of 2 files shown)


jannic (approved):
I can't claim that I understood this in detail, but the explanation sounds sensible, and the code seems to match the explanation. So probably it's either correct or it was written by ChatGPT :-)

PR #111, opened 2023-12-17
Support making directories

thejpster:
Now you can make a directory inside a directory.

### DIFF
--- a/src/volume_mgr.rs
+++ b/src/volume_mgr.rs
@@ -5,7 +5,7 @@
 use byteorder::{ByteOrder, LittleEndian};
 use core::convert::TryFrom;
 
-use crate::fat::{self, BlockCache, RESERVED_ENTRIES};
+use crate::fat::{self, BlockCache, FatType, OnDiskDirEntry, RESERVED_ENTRIES};
 
 use crate::filesystem::{
     Attributes, ClusterId, DirEntry, DirectoryInfo, FileInfo, Mode, RawDirectory, RawFile,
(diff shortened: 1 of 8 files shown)


src/volume_mgr.rs
```diff
@@ -866,6 +864,146 @@ where
...
+                    .copy_from_slice(&dot_dot_entry_in_child.serialize(fat_type)[..]);
+
+                self.block_device
+                    .write(&blocks, new_dir_start_block)
+                    .map_err(Error::DeviceError)?;
+
+                // Now zero the rest of the cluster
+                for b in blocks[0].iter_mut() {
+                    *b = 0;
+                }
```
jannic, line 992: Shouldn't this be the other way around?
```suggestion
                // Now zero the rest of the cluster
                for b in blocks[0].iter_mut() {
                    *b = 0;
                }
                
                self.block_device
                    .write(&blocks, new_dir_start_block)
                    .map_err(Error::DeviceError)?;
```

thejpster: You I that it wouldn't write the two directory entries I just carefully put into that block. Besides, it was zero initialised.

Having written the first block in the cluster, I need to write zeros to the rest of the cluster (which is up to 32 blocks long). I already had a block on the stack so I zeroed it. I could make a second block, but that probably just chews up more stack space.

jannic: You are of course right. I was dumb and didn't read that code carefully, and somehow thought "Now zero the rest of the cluster" was about clearing the rest of `blocks[0]` before it gets written to the card. But that doesn't make sense. As you said, it was zero initialized from the beginning anyway.

jannic (approved):

PR #112, opened 2023-12-19
Enhanced shell example

thejpster:
The shell example now supports some relative paths and some new commands.

### DIFF
--- a/examples/shell.rs
+++ b/examples/shell.rs
@@ -1,6 +1,73 @@
 //! A simple shell demo for embedded-sdmmc
 //!
-//! Presents a basic command prompt which implements some basic MS-DOS style shell commands.
+//! Presents a basic command prompt which implements some basic MS-DOS style
+//! shell commands.
+//!
+//! Note that `embedded_sdmmc` itself does not care about 'paths' - only
+//! accessing files and directories on on disk, relative to some previously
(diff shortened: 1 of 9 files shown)


examples/shell.rs
```diff
@@ -30,134 +195,305 @@ impl Context {
...
+    /// Resolves a filename.
+    ///
+    /// Converts a string path into a directory handle and a name within that
+    /// directory (that may or may not exist).
+    ///
+    /// * Bare names (no leading `.`, `/` or `N:/`) are mapped to the current
+    ///   directory in the current volume.
+    /// * Relative names, like `../SOMEDIR/SOMEFILE` or `./SOMEDIR/SOMEFILE`, traverse
+    ///   starting at the current volume and directory.
+    /// * Absolute, like `1:/SOMEDIR/SOMEFILE` start at the given volume.
```
jannic, line 457: Is it intentional to use 1 as a volume label in this example?
Also, a few lines above, N is used as a volume label, but the current implementation only supports A-D.

thejpster: No, that's a left over. Good catch!

thejpster:
@jannic  is this good to go after those changes?

jannic (approved):

Issue #98, opened 2023-09-22
Add a flush API


thejpster:
An open file is risky because the data is still in RAM. Add a flush API which is like a close and a reopen.

peterkrull:
I just started using this crate. I intend to use it for high-rate logging on an aerial drone, where the probability of a crash and the log files being needed are highly correlated. So ensuring data is saved, even in the event the SD card is flung loose or instant loss of power, would be nice. I presume this flushing API would be a way to accomplish that?

thejpster:
Yes. It would be like closing the file and re-opening it. Basically you just need to update the directory entry to have the correct file size.

But also that sounds like a use-case where your data is important to you, and I only really wrote this crate to suit my own whimsical needs - so please do your own testing and verification before trusting it with anything.

peterkrull:
Thanks for the quick response. I briefly tried experimenting with dropping the `File` handle and acquiring a new one with `Directory::open_file_in_dir(..)` to force it to update the directory entry. This caused a significant slowdown in throughput if it was done too often, but it did like I was more likely to have a file if I yanked the SD card while my program was running. I will keeping testing and looking around to figure out what I need, but the ability to flush would definitely be a nice feature!

PR #121, opened 2024-03-25
Adds flush API

peterkrull:
Adds the ability to manually flush/commit written data to the directory entry, as discussed in #98.

### DIFF
--- a/src/filesystem/files.rs
+++ b/src/filesystem/files.rs
@@ -1,6 +1,6 @@
 use crate::{
     filesystem::{ClusterId, DirEntry, SearchId},
-    RawVolume, VolumeManager,
+    Error, RawVolume, VolumeManager,
 };
 
 /// Represents an open file on disk.
@@ -122,6 +122,11 @@ where
         let f = self.raw_file;
         core::mem::forget(self);
         f
+    }
+
+    /// Flush any written data by updating the directory entry.
+    pub fn flush(&mut self) -> Result<(), Error<D::Error>> {
+        self.volume_mgr.flush_file(self.raw_file)
     }
 }
 
--- a/src/volume_mgr.rs
+++ b/src/volume_mgr.rs
@@ -773,36 +773,16 @@ where
         Ok(())
     }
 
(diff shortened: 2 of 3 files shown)


src/filesystem/files.rs
```diff
@@ -138,6 +138,18 @@ where
...
 }
 
+impl<'a, D, T, const MAX_DIRS: usize, const MAX_FILES: usize, const MAX_VOLUMES: usize>
+    File<'a, D, T, MAX_DIRS, MAX_FILES, MAX_VOLUMES>
+where
+    D: crate::BlockDevice,
+    T: crate::TimeSource,
+{
+    /// Flush any written data by updating the directory entry.
+    pub fn flush(&mut self) -> Result<(), Error<D::Error>> {
```
thejpster, line 148: Can we put this function in this existing impl block above?

src/volume_mgr.rs
```diff
@@ -775,6 +775,13 @@ where
 
     /// Close a file with the given full path.
     pub fn close_file(&mut self, file: RawFile) -> Result<(), Error<D::Error>> {
+        let file_idx = self.flush_file(file)?;
+        self.open_files.swap_remove(file_idx);
+        Ok(())
+    }
+
+    /// Flush (update the entry) for a file with the given full path.
```
thejpster, line 783: change "given full path" to "given raw file handle".

thejpster: It also doesn't make sense to return the internal index here. Maybe make this a private function, and have the public `close_file` and `flush_file` call it?

thejpster (commented):
In general, this looks good. Just a few minor notes as I read through.

I wonder also if we can a test which creates a file, writes a few bytes, flushes it, and then checks the directory entry is updated with the new file size.

peterkrull:
@thejpster I have implemented your suggestions

thejpster:
Just a cargo fmt and I think we're good to go.

thejpster (approved):

Issue #122, opened 2024-03-25
In case of early ejection of SD card, allow "closing" `RawFile` handle anyway


peterkrull:
It seems that there is no nice way to get rid of a RawFile handle if for example the SD card is ejected, since calls to `crate::VolumeManager::close_file` will result in a returned error before the file id is removed from the `open_files` list.

In my testing, I have implemented the following:

```rust
if volume_mgr.write(file, data.as_slice()).is_err() {
    volume_mgr.destroy_file(file).unwrap();
    volume_mgr.close_dir(dir).unwrap();
    volume_mgr.close_volume(volume).unwrap();

    continue 'setup; // Loop back and retry connecting to card
}
```

where 
```rust
/// Destroy a file for the given RawFile, without updating its entry.
/// This can be useful if the SD card is suddenly ejected, and one must
/// notify the VolumeManager that we are dropping the handle.
pub fn destroy_file(&mut self, file: RawFile) -> Result<(), Error<D::Error>> {
    let file_idx = self.get_file_by_id(file)?;
    self.open_files.swap_remove(file_idx);
    Ok(())
}
```
Which works nicely. Another option would be to modify the `close_file` method to first get the index, remove from the list, and then try to update the entry by flusing (as implemented in #121):

```rust
/// Close a file with the given full path.
pub fn close_file(&mut self, file: RawFile) -> Result<(), Error<D::Error>> {
    let file_idx = self.get_file_by_id(file)?;
    self.open_files.swap_remove(file_idx);
    self.flush(file)?;
    Ok(())
}
```

thejpster:
I think `close` should always close a file, regardless of whether it was able to flush it.

peterkrull:
I think the simplest solution is to simply not short-circuit the file-flush, then remove the file index, then return the result of the file-flush

peterkrull:
Closing in favor of #123

PR #123, opened 2024-03-25
Always closing (and more)

peterkrull:
Answering my own issue #122 with this one.

When calling `VolumeManager::close_file(..)` one would expect the file index to actually be removed from the volume manager, even if flushing the data to the card fails. This is currently not the case due to any errors during flushing short-circuiting the error. This PR addresses that.

I have also taken some the liberties in turning some for-loop searches into iterators, and removed panicking behavior from the `impl Drop`s on volumes, directories and files.

### DIFF
--- a/src/filesystem/files.rs
+++ b/src/filesystem/files.rs
@@ -39,9 +39,12 @@ impl RawFile {
 
 /// Represents an open file on disk.
 ///
-/// If you drop a value of this type, it closes the file automatically. However,
-/// it holds a mutable reference to its parent `VolumeManager`, which restricts
-/// which operations you can perform.
+/// In contrast to a `RawFile`, a `File`  holds a mutable reference to its
+/// parent `VolumeManager`, which restricts which operations you can perform.
+///
+/// If you drop a value of this type, it closes the file automatically, and but
+/// error that may occur will be ignored. To handle potential errors, use
+/// the [`File::close`] method.
 pub struct File<'a, D, T, const MAX_DIRS: usize, const MAX_FILES: usize, const MAX_VOLUMES: usize>
 where
     D: crate::BlockDevice,
@@ -128,6 +131,16 @@ where
     pub fn flush(&mut self) -> Result<(), Error<D::Error>> {
         self.volume_mgr.flush_file(self.raw_file)
     }
+
+    /// Consume the `File` handle and close it. The behavior of this is similar
+    /// to using [`core::mem::drop`] or letting the `File` go out of scope,
+    /// except this lets the user handle any errors that may occur in the process,
+    /// whereas when using drop, any errors will be discarded silently.
+    pub fn close(self) -> Result<(), Error<D::Error>> {
+        let result = self.volume_mgr.close_file(self.raw_file);
+        core::mem::forget(self);
+        result
+    }
 }
 
 impl<'a, D, T, const MAX_DIRS: usize, const MAX_FILES: usize, const MAX_VOLUMES: usize> Drop
@@ -137,9 +150,7 @@ where
     T: crate::TimeSource,
 {
     fn drop(&mut self) {
-        self.volume_mgr
-            .close_file(self.raw_file)
-            .expect("Failed to close file");
+        _ = self.volume_mgr.close_file(self.raw_file);
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -239,9 +239,12 @@ impl RawVolume {
 
 /// Represents an open volume on disk.
 ///
-/// If you drop a value of this type, it closes the volume automatically. However,
-/// it holds a mutable reference to its parent `VolumeManager`, which restricts
-/// which operations you can perform.
+/// In contrast to a `RawVolume`, a `Volume` holds a mutable reference to its
+/// parent `VolumeManager`, which restricts which operations you can perform.
+///
+/// If you drop a value of this type, it closes the volume automatically, but
+/// any error that may occur will be ignored. To handle potential errors, use
+/// the [`Volume::close`] method.
 pub struct Volume<'a, D, T, const MAX_DIRS: usize, const MAX_FILES: usize, const MAX_VOLUMES: usize>
 where
     D: crate::BlockDevice,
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -294,9 +294,7 @@ where
     T: crate::TimeSource,
 {
     fn drop(&mut self) {
-        self.volume_mgr
-            .close_volume(self.raw_volume)
-            .expect("Failed to close volume");
+        _ = self.volume_mgr.close_volume(self.raw_volume)
```
jonathanpallant, line 297: I’m not sure about silently hiding these errors on drop. I think I would prefer to offer a non-panicking API and leave Drop as panicking. But I’m happy to be persuaded otherwise.

src/volume_mgr.rs
```diff
@@ -775,14 +775,36 @@ where
 
     /// Close a file with the given raw file handle.
     pub fn close_file(&mut self, file: RawFile) -> Result<(), Error<D::Error>> {
-        let file_idx = self.flush_file_get_index(file)?;
+        let flush_result = self.flush_file(file);
+        let file_idx = self.get_file_by_id(file)?;
```
jonathanpallant, line 779: It’s unfortunate to have to find the file index twice.

src/volume_mgr.rs
```diff
@@ -993,30 +1015,30 @@ where
     }
 
     fn get_volume_by_id(&self, volume: RawVolume) -> Result<usize, Error<D::Error>> {
-        for (idx, v) in self.open_volumes.iter().enumerate() {
-            if v.volume_id == volume {
-                return Ok(idx);
-            }
-        }
-        Err(Error::BadHandle)
+        self.open_volumes
```
jonathanpallant, line 1018: I’m not sure this is a big enough improvement to be worth the change?

src/lib.rs
peterkrull: I guess the non-panicking drop would just be to turn the `File` into a `RawFile` and close using the volume manager, or make a dedicated method for that on the `File` itself. My problem is just that if some part of a program relies on the `Drop` to close the file, then if the card is ejected or unresponsive, that will cause a panic. That to me is quite scary, if lets say, I am implementing a system where the logging module is very much *not* mission critical, but it still causes a crash on a bad connection.

I think the documentation for the `drop(..)` method explains the reason for avoiding panics being a general recommendation pretty well. But I also think think it is important to consider what behavior is desired/expected for the library.

Does an error during a close cause the program to be in an undefined or unrecoverable state? If not, I'd say the best-effort close is better than a panic. Perhaps it is only really relevant to consider the implications of the `File` drop, since causing the `Volume` or `Directory` drop to panic is not really possible as far as I can tell.

I just found it head-scratching that the program panicked when I first tried to test yanking the card during operation. I expected some data to not be flushed (before my previous commit), but not a panic. At the very least it should be loud and clear in the documentation for the `File` type that they will panic if they are not able to drop. Especially since actually encountering the panic during testing is not guaranteed if the card disconnecting is a very rate event.

src/volume_mgr.rs
peterkrull: Yeah I thought about that. But, I also think the previous flush_file_get_index was a bit of an awkward dual-purpose function. This implementation is ever so slightly less performant at closing a file, but the functions and their purpose are clearer.

src/volume_mgr.rs
peterkrull: Yeah not really an improvement per se, the assembly is nearly identical between the two. More a style preference, on for-loops vs iterators. No problem reverting this

src/volume_mgr.rs
jonathanpallant: This is persuasive

src/lib.rs
jonathanpallant: I was thinking that "Silent failures are bad", and drop can't do anything but panic on error. But std::fs::File says:

> Errors detected on closing are ignored by the implementation of Drop.

src/filesystem/files.rs
```diff
@@ -128,6 +128,15 @@ where
     pub fn flush(&mut self) -> Result<(), Error<D::Error>> {
         self.volume_mgr.flush_file(self.raw_file)
     }
+
+    /// Consume the `File` handle and close it. The behavior of this is similar
+    /// to using [`core::mem::drop`] or letting the `File` go out of scope,
+    /// except if the file fails to close properly, this returns an error,
+    /// whereas the automatic drop will panic. This method is therefore hihgly
```
jonathanpallant, line 135: typo

src/volume_mgr.rs
jonathanpallant: I think that would make sense

jonathanpallant (dismissed):

src/lib.rs
peterkrull: Comparing to something similar in std library makes sense! And I think it goes with the idea of "do what the user would expect". Should we do it for File, Volume and Directory?

src/filesystem/files.rs
peterkrull: If you agreed that not panicking on drop was the way to go, I would remove the part about panicking here, and instead do something like:

```
/// Consume the `File` handle and close it. The behavior of this is similar
/// to using [`core::mem::drop`] or letting the `File` go out of scope,
/// except this lets the user handle any errors that may occur in the process,
/// whereas when using drop, any errors will be discarded silently.
```

src/filesystem/files.rs
```diff
@@ -128,6 +128,14 @@ where
...
         self.volume_mgr.flush_file(self.raw_file)
     }
+
+    /// Consume the `File` handle and close it. The behavior of this is similar
+    /// to using [`core::mem::drop`] or letting the `File` go out of scope,
+    /// except this lets the user handle any errors that may occur in the process,
+    /// whereas when using drop, any errors will be discarded silently.
+    pub fn close(self) -> Result<(), Error<D::Error>> {
+        self.volume_mgr.close_file(self.raw_file)
+    }
```
thejpster, line 138: I think you need to forget the File here, to prevent the Drop trait from running. Otherwise it'll try and close it twice, and fail the second time - which just seems wasteful.

src/lib.rs
thejpster: Being consistent sounds good, if the change is well documented.

thejpster (changes requested):

thejpster:
Sorry, spotted a few more things on this pass. And would you mind squashing the branch - we've got some commits that are now redundant.

thejpster (approved):

thejpster:
Looks good to me - thank you!

PR #132, opened 2024-05-30
implement FAT table write/read and setting the volume dirty status

Murmele:
The dirty flag will be set when opening the volume to indicate that the FAT table might be out of date. In case of an unexpected close without unmount in the next mount it is visible that the FAT table is outdated and can be reread by the operating system (https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system). Without this, the file must always closed before removing the sd card because otherwise the FAT table is outdated and so the files are not visible when mounting the sd card on a computer. For embedded systems it is quite likely that unexpect end can happen and always closing the file after a single write operation is a performance issue. For readonly this dirty flag is not required and therefore it will not set and reset when opening/closing a volume. This is same as the linux kernel is doing.

https://unix.stackexchange.com/questions/230181/why-does-linux-mark-fat-as-dirty-simply-due-to-mounting-it Linux 

kernel is doing it during mount, so there is no need to ckeck it every time during a new write

### DIFF
--- a/src/volume_mgr.rs
+++ b/src/volume_mgr.rs
@@ -12,8 +12,8 @@ use crate::filesystem::{
     SearchIdGenerator, TimeSource, ToShortFileName, MAX_FILE_SIZE,
 };
 use crate::{
-    debug, Block, BlockCount, BlockDevice, BlockIdx, Error, RawVolume, ShortFileName, Volume,
-    VolumeIdx, VolumeInfo, VolumeType, PARTITION_ID_FAT16, PARTITION_ID_FAT16_LBA,
+    debug, Block, BlockCount, BlockDevice, BlockIdx, Error, FatVolume, RawVolume, ShortFileName,
+    Volume, VolumeIdx, VolumeInfo, VolumeType, PARTITION_ID_FAT16, PARTITION_ID_FAT16_LBA,
     PARTITION_ID_FAT32_CHS_LBA, PARTITION_ID_FAT32_LBA,
 };
 use heapless::Vec;
@@ -102,8 +102,26 @@ where
     pub fn open_volume(
         &mut self,
         volume_idx: VolumeIdx,
+        read_only: bool,
     ) -> Result<Volume<D, T, MAX_DIRS, MAX_FILES, MAX_VOLUMES>, Error<D::Error>> {
-        let v = self.open_raw_volume(volume_idx)?;
+        return self._open_volume(volume_idx, false);
+    }
+
+    /// Get a volume (or partition) based on entries in the Master Boot Record.
+    ///
+    /// We do not support GUID Partition Table disks. Nor do we support any
+    /// concept of drive letters - that is for a higher layer to handle.
+    fn _open_volume(
+        &mut self,
+        volume_idx: VolumeIdx,
+        read_only: bool,
+    ) -> Result<Volume<D, T, MAX_DIRS, MAX_FILES, MAX_VOLUMES>, Error<D::Error>> {
+        let v = self.open_raw_volume(volume_idx, read_only)?;
+        if !read_only {
+            let idx = self.get_volume_by_id(v)?;
+            let VolumeType::Fat(volume_type) = &self.open_volumes[idx].volume_type;
+            self.set_volume_status_dirty(volume_type, true)?;
+        }
         Ok(v.to_volume(self))
     }
 
@@ -114,7 +132,11 @@ where
     ///
     /// This function gives you a `RawVolume` and you must close the volume by
     /// calling `VolumeManager::close_volume`.
-    pub fn open_raw_volume(&mut self, volume_idx: VolumeIdx) -> Result<RawVolume, Error<D::Error>> {
+    pub fn open_raw_volume(
+        &mut self,
+        volume_idx: VolumeIdx,
+        read_only: bool,
+    ) -> Result<RawVolume, Error<D::Error>> {
         const PARTITION1_START: usize = 446;
         const PARTITION2_START: usize = PARTITION1_START + PARTITION_INFO_LENGTH;
         const PARTITION3_START: usize = PARTITION2_START + PARTITION_INFO_LENGTH;
@@ -192,6 +214,7 @@ where
                     volume_id: id,
                     idx: volume_idx,
                     volume_type: volume,
+                    read_only: read_only,
                 };
                 // We already checked for space
                 self.open_volumes.push(info).unwrap();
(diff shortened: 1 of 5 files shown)


thejpster:
It appears the tests are failing - possible because the API changed?

src/volume_mgr.rs
```diff
@@ -114,7 +143,11 @@ where
     ///
     /// This function gives you a `RawVolume` and you must close the volume by
     /// calling `VolumeManager::close_volume`.
-    pub fn open_raw_volume(&mut self, volume_idx: VolumeIdx) -> Result<RawVolume, Error<D::Error>> {
+    pub fn open_raw_volume(
+        &mut self,
+        volume_idx: VolumeIdx,
+        read_only: bool,
```
thejpster, line 149: I think I'd prefer an `enum VolumeOpenMode { ReadOnly, ReadWrite }`, otherwise the caller writes:

`thing.open_raw_volume(VolumeIdx(0), true)`

and it's totally unclear what `true` means from the context.

src/volume_mgr.rs
```diff
@@ -103,7 +103,36 @@ where
...
+        volume_idx: VolumeIdx,
+    ) -> Result<Volume<D, T, MAX_DIRS, MAX_FILES, MAX_VOLUMES>, Error<D::Error>> {
+        return self._open_volume(volume_idx, true);
+    }
+
+    /// Get a volume (or partition) based on entries in the Master Boot Record.
+    ///
+    /// We do not support GUID Partition Table disks. Nor do we support any
+    /// concept of drive letters - that is for a higher layer to handle.
+    fn _open_volume(
```
thejpster, line 125: We added an argument to `open_raw_volume`, so should we just add an argument to `open_volume`, and not have the two extra methods? Not sure.

src/volume_mgr.rs
```diff
@@ -319,13 +353,55 @@ where
...
+            self.block_device.write(&blocks, fat_table1_start)?;
+            // Synchronize also backup fat table
+            if volume.fat_nums == 2 {
+                self.block_device
+                    .write(&blocks, fat_table1_start + volume.fat_size)?
+            }
+        }
+        Ok(())
+    }
+    #[cfg(test)]
```
thejpster, line 392: Tests go in `mod test {}`.

src/volume_mgr.rs
```diff
@@ -319,13 +353,55 @@ where
                 return Err(Error::VolumeStillInUse);
             }
         }
-
         let volume_idx = self.get_volume_by_id(volume)?;
+        if !self.open_volumes[volume_idx].read_only {
+            let VolumeType::Fat(volume_type) = &self.open_volumes[volume_idx].volume_type;
```
thejpster, line 358: I've tried to write this code on the assumption that a non-FAT filesystem will be added in the future. The irrefutable match here will break when that happens - I think `set_volume_status_dirty` doesn't need `volume_type` as an argument - it can just work it out.

src/volume_mgr.rs
```diff
@@ -319,13 +353,55 @@ where
...
+
         self.open_volumes.swap_remove(volume_idx);
 
         Ok(())
     }
 
+    /// Sets the volume status dirty to dirty if true, to not dirty if false
+    fn set_volume_status_dirty(
+        &self,
+        volume: &FatVolume,
```
thejpster, line 370: I think this should take `volume_idx` and handle the different types of volume internally. It should also skip updating the FAT if the FAT is already marked as dirty.

src/volume_mgr.rs
```diff
@@ -103,7 +103,36 @@ where
...
+    fn _open_volume(
+        &mut self,
+        volume_idx: VolumeIdx,
+        read_only: bool,
+    ) -> Result<Volume<D, T, MAX_DIRS, MAX_FILES, MAX_VOLUMES>, Error<D::Error>> {
+        let v = self.open_raw_volume(volume_idx, read_only)?;
+        if !read_only {
+            let idx = self.get_volume_by_id(v)?;
+            let VolumeType::Fat(volume_type) = &self.open_volumes[idx].volume_type;
+            self.set_volume_status_dirty(volume_type, true)?;
```
thejpster, line 134: I don't think you need to mark it as dirty until the first write occurs?

thejpster (commented):
I like the idea - left a few notes.

src/volume_mgr.rs
thejpster: Ah, I read the comment thread from the kernel. Yeah, I can see the value in marking it dirty on first mount.

Note to self - the Neotron OS will need changing to ensure the volume is unmounted when the last file is closed.

src/volume_mgr.rs
Murmele: Good point thanks!

Murmele:
> It appears the tests are failing - possible because the API changed?

I will have a look into it next days

src/volume_mgr.rs
Murmele: But this flag does not fix my problem haha, I have to update somehow the fat, at least Linux does not check the fat and repairs it if needed. I thought when the bit is cleared (is dirty) during mount it will be "repaired" and the FAT table gets update, but this is not the case

src/volume_mgr.rs
Murmele: changed in 2fd18cbaa92ca31ceac2922de05e50e3511ca88b

src/volume_mgr.rs
Murmele: I made it like this, so it is downward compatible

src/volume_mgr.rs
Murmele: changed in 9d6865a75355be25cd1b3bbc5597d29d2862591c

src/volume_mgr.rs
Murmele: fixed in 56a07fec25eb9af5f54b90b6e9d49f275962b149

src/volume_mgr.rs
thejpster: Then should we keep open_raw_volume backwards compatible too?

src/volume_mgr.rs
Murmele: I changed this

src/volume_mgr.rs
Murmele: Is there a need for? I though because it will be used mostly internally so the impact is lower?

thejpster: it's a public API, which I use in my projects.

Murmele: I can change also to the same structure as for open_volume or do you prefer the argument way for both?

src/volume_mgr.rs
thejpster: I don't think Linux will run an `fsck` on a block device unless you ask it to. It is Windows that prompts to run Scandisk if it things the volume wasn't cleanly shut down.

src/volume_mgr.rs
thejpster: I think I prefer it with the argument.

thejpster:
So we already have functions for updating the FAT: `FatVolume::update_fat`. Could we just create some new `ClusterId` variants which represent clean/dirty volumes and use the `update_fat` function to write them to ClusterIdx 1?

src/volume_mgr.rs
Murmele: Perfect then I will change it also for the open_volume

Murmele: fixed in 70a5fdbe7d9aa69273baa77fc23012ea9c34666d

Murmele:
> So we already have functions for updating the FAT: `FatVolume::update_fat`. Could we just create some new `ClusterId` variants which represent clean/dirty volumes and use the `update_fat` function to write them to ClusterIdx 1?

My idea was now with this flag, after a restart of my device I check this flag and if dirty I wanna update somehow the FAT to show the correct information. I have to check if I can get there somehow. Sorry right now I am not that deep into the topic to understand your idea. I will think about and doing some research.

In my case the sd card never leaves the device but at the end over usb storage access shall be established, so I can easily update the FAT at that point

Murmele:
I fixed the tests

thejpster:
> So we already have functions for updating the FAT: `FatVolume::update_fat`. Could we just create some new `ClusterId` variants which represent clean/dirty volumes and use the `update_fat` function to write them to ClusterIdx 1?

So I like the read-only and read-write changes. If they were isolated I'd happily merge them.

But to return to my earlier comment and perhaps clarify - I don't think we need a whole new module for writing a couple of FAT entries to indicate dirty/clean, when we already have functions which write FAT entries for the purposes of tracking cluster allocations for files. Either all the FAT writing code needs to go in to the new module, or the clean/dirty bits need to be set with the existing function.

Murmele:
> > So we already have functions for updating the FAT: `FatVolume::update_fat`. Could we just create some new `ClusterId` variants which represent clean/dirty volumes and use the `update_fat` function to write them to ClusterIdx 1?
> 
> So I like the read-only and read-write changes. If they were isolated I'd happily merge them.
> 
> But to return to my earlier comment and perhaps clarify - I don't think we need a whole new module for writing a couple of FAT entries to indicate dirty/clean, when we already have functions which write FAT entries for the purposes of tracking cluster allocations for files. Either all the FAT writing code needs to go in to the new module, or the clean/dirty bits need to be set with the existing function.

I will have a look into it. Thanks!

PR #136, opened 2024-07-06
Use SpiDevice to control the chip select.

### LINKED ISSUE #126: should use SpiBus, not SpiDevice
https://github.com/rust-embedded-community/embedded-sdmmc-rs/blob/710bd34253d453de4045c7307d1d286ccc1972ed/src/sdcard/mod.rs#L61-L67

If you take a CS pin directly, you have to use SpiBus, not SpiDevice. 

Using SpiDevice with a dummy CS pin breaks bus sharing. (the bus looks "free" to the mutex between transactions, so it might grant the mutex to some other driver, which will break because this lib still has the CS pin asserted)

embedded-hal docs explicitly say not to do this https://docs.rs/embedded-hal/latest/embedded_hal/spi/index.html#for-driver-authors

thejpster:
Deletes the `CS` type param and associated field. Now we just rely on the `SpiDevice` we were given to control the chip select. This means we can't do the 74 cycles of card warm-up anymore - the caller will have to work out how to do that in a way that doesn't break the bus-sharing rules of `SpiDevice`.

Closes #126

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ log = {version = "0.4", default-features = false, optional = true}
 
 [dev-dependencies]
 chrono = "0.4"
-embedded-hal-bus = "0.1.0"
+embedded-hal-bus = "0.2.0"
 env_logger = "0.10.0"
 flate2 = "1.0"
 hex-literal = "0.4.1"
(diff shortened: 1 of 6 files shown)


thejpster:
This needs testing on a bunch of cards to ensure it works OK.

I tried a Kingston 16GB micro SDHC, and it was OK provided I did give it the 74 start-up clock cycles in my application. If I didn't do that, it returned `0` from a `CMD0`, and so the card was rejected by embedded-sdmmc as not working.

thejpster:
Wilfried tried it for me and it seemed to work on a few cards he had.

ryan-summers (approved):

PR #143, opened 2024-10-06
Use a RefCell to provide interior mutability for VolumeManager.

### LINKED ISSUE #79: Explore interior mutability
Now that `critical-section` is a thing, maybe we can consider interior mutability:

* `Volume` can hold an `&VolumeManager`
* `Directory` and `File` can hold an `&Volume`

```rust
let volume = volume_mgr.get_volume(VolumeIdx(0))?;
let root_dir = volume.open_root_dir()?;
let file = root_dir.open("HELLO.TXT")?;
let mut buffer = [0u8; 64];
let contents = file.read_to_buffer(&mut buffer)?;
```

thejpster:
VolumeManager now contains a RefCell, so all the methods can be `&self`. This also means you can have multiple smart handles usable at the same time.

Note that VolumeManager is !Sync, so this is just about having multiple `File` objects active at one time.

Closes #79 (finally!)

### DIFF
--- a/src/volume_mgr.rs
+++ b/src/volume_mgr.rs
@@ -218,8 +218,9 @@ where
     /// You can then read the directory entries with `iterate_dir`, or you can
     /// use `open_file_in_dir`.
     pub fn open_root_dir(&self, volume: RawVolume) -> Result<RawDirectory, Error<D::Error>> {
-        // Opening a root directory twice is OK
+        debug!("Opening root on {:?}", volume);
 
+        // Opening a root directory twice is OK
         let mut data = self.data.try_borrow_mut().map_err(|_| Error::LockError)?;
 
         let directory_id = RawDirectory(data.id_generator.generate());
@@ -232,6 +233,8 @@ where
         data.open_dirs
             .push(dir_info)
             .map_err(|_| Error::TooManyOpenDirs)?;
+
+        debug!("Opened root on {:?}, got {:?}", volume, directory_id);
 
         Ok(directory_id)
     }
@@ -312,6 +315,7 @@ where
     /// Close a directory. You cannot perform operations on an open directory
     /// and so must close it if you want to do something with it.
     pub fn close_dir(&self, directory: RawDirectory) -> Result<(), Error<D::Error>> {
+        debug!("Closing {:?}", directory);
         let mut data = self.data.try_borrow_mut().map_err(|_| Error::LockError)?;
 
         for (idx, info) in data.open_dirs.iter().enumerate() {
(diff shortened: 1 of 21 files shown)


src/volume_mgr.rs
```diff
@@ -130,11 +135,13 @@ where
         const PARTITION_INFO_LBA_START_INDEX: usize = 8;
         const PARTITION_INFO_NUM_BLOCKS_INDEX: usize = 12;
 
-        if self.open_volumes.is_full() {
+        let mut data = self.data.borrow_mut();
```
jannic, line 138: `borrow_mut` may panic.
I'd prefer if this used `try_borrow_mut` and returned an appropriate error in case the RefCell is already borrowed.

(Not only here, but I won't copy the same comment to all calls to `borrow_mut`.)

jannic (approved):
I'd prefer the change to try_borrow_mut, unless I'm missing some reason why that wouldn't be easily possible.

Otherwise, LGTM.

src/volume_mgr.rs
thejpster: Wouldn't a double-locking indicate a fault in the filesystem code as opposed to anything the user can do anything about?

thejpster: I added a commit with the requested change. Let me know if you do indeed prefer it.

jannic: Perhaps I'm misunderstanding something? With this pull request, isn't it possible to share a `&VolumeManager` between threads (cores, interrupt context, whatever), so calls to these public functions can happen concurrently?
How does it require a fault in the filesystem code?

thejpster: I thought it was !Sync. That was certainly the intention.

jannic: Of course! That's what I was missing: I looked at the docs for the crate, and noticed that VolumeManager implements `Send` and `Sync`. But that will no longer be true with this change, as the contained `RefCell` will prevent the auto trait `Sync` from being implemented.

Let me have another look at the code, but indeed I guess that makes the error handling unnecessary.

thejpster: Yeah, RefCell is [expressly !Sync](https://doc.rust-lang.org/stable/std/cell/struct.RefCell.html#impl-Sync-for-RefCell%3CT%3E). So you can't pass a reference to a VolumeManager between threads.

thejpster:
I rebased.

src/volume_mgr.rs
jannic: Yes, I know - I just didn't make the connection that the auto-trait gets inherited, so `data` being `!Sync` will automatically cause `VolumeManager` to also be `!Sync`.

Indeed, considering that, currently I don't see a way for multiple borrows to happen at the same time.
The only function that both borrows `data` and calls another function that also borrows `data` is `close_file`, and that one calls `flush_file` before borrowing `data`, so it should be fine.

And I agree, if a future change of the crate violates that property and somehow does recursive calls to `borrow_mut` / `borrow`, that would be a severe bug that could justify a panic.

I'm a bit indecisive whether a panic is _always_ the right response. The firmware may be prepared to continue running after filesystem errors anyway (the SD card may be a non-essential part). In that case, returning a `LockError` could convert a catastrophic (but obvious) bug into a minor inconvenience. At the same time it would hide the underlying bug, and the end-user may not get any useful indication why the appliance stops accessing the file system.

Overall, I'm fine with both variants. With the `shell` example, I don't see a size difference.
Do you prefer panicking version?

thejpster: > Indeed, considering that, currently I don't see a way for multiple borrows to happen at the same time.

I think that maybe you could try and open a file from the iterate closure.

thejpster: Yes I can confirm that trying to do anything with the VolumeManager from inside the iterate closure produces a lock error. So returning the error is better than panicking.

thejpster:
Rust is brilliant. I had very little trouble chopping up this codebase and hacking on that shell example to use the auto-closing `Directory` objects. If it compiles, it has basically worked.

thejpster:
Rebased again 😢 

Make small PRs, they said. It's easier to review, they said.

jannic:
> Make small PRs, they said. It's easier to review, they said.

Don't worry. It _is_ easier to review!

src/volume_mgr.rs
jannic: Good catch, I didn't notice that possibility. Perhaps in this case you want to reconsider the doc comment on LockError? "This is a bug in the filesystem. Please open an issue."

jannic (dismissed):

thejpster:
I wonder if I should turn off the thing that dismisses reviews as stale when a commit is added.

thejpster:
OK, all rebased and notes added about double locking. I think this monster is good to go.

src/volume_mgr.rs
thejpster: Added some better notes, and a warning on the relevant closure-based APIs.

jannic (approved):

PR #144, opened 2024-10-06
Better volume label support

thejpster:
Now you can ask for the label of a volume. The `shell` example also now prints the volume label on start-up.

```console
$ cargo run --example shell --release -- ./disk.img
    Finished `release` profile [optimized] target(s) in 0.03s
     Running `target/release/examples/shell ./disk.img`
Opening './disk.img'...
Volume # A: found, label: Some(VolumeName("TEST FAT16"))
Volume # B: found, label: Some(VolumeName("TEST FAT32"))
Failed to open volume 2: FormatError("Partition type not supported")
Failed to open volume 3: FormatError("Partition type not supported")
A:/> 
```

### DIFF
--- a/src/fat/bpb.rs
+++ b/src/fat/bpb.rs
@@ -85,11 +85,13 @@ impl<'a> Bpb<'a> {
     // FAT16/FAT32 functions
 
     /// Get the Volume Label string for this volume
-    pub fn volume_label(&self) -> &[u8] {
+    pub fn volume_label(&self) -> [u8; 11] {
+        let mut result = [0u8; 11];
         match self.fat_type {
-            FatType::Fat16 => &self.data[43..=53],
-            FatType::Fat32 => &self.data[71..=81],
+            FatType::Fat16 => result.copy_from_slice(&self.data[43..=53]),
+            FatType::Fat32 => result.copy_from_slice(&self.data[71..=81]),
         }
+        result
     }
 
     // FAT32 only functions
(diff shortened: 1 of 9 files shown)


thejpster:
Closes #145

src/fat/bpb.rs
```diff
@@ -85,23 +85,23 @@ impl<'a> Bpb<'a> {
...
-    pub fn volume_label(&self) -> &[u8] {
-        if self.fat_type != FatType::Fat32 {
-            &self.data[43..=53]
-        } else {
-            &self.data[71..=81]
+    pub fn volume_label(&self) -> [u8; 11] {
+        let mut result = [0u8; 11];
+        match self.fat_type {
+            FatType::Fat16 => result.copy_from_slice(&self.data[43..=53]),
+            FatType::Fat32 => result.copy_from_slice(&self.data[71..=81]),
```
jannic, line 92: According to https://en.wikipedia.org/wiki/Design_of_the_FAT_file_system, this field is not available on Fat32 "if the signature at offset 0x042 is set to 0x28".

Perhaps that should be checked here?

jannic (approved):

thejpster:
Rebased

src/fat/bpb.rs
thejpster: > Most FAT32 file system implementations do not support an alternative signature of 0x28

> Should be 0x29 to indicate that an EBPB with the following 3 entries exists (since OS/2 1.2 and DOS 4.0). Can be 0x28 on some OS/2 1.0-1.1 and PC DOS 3.4 disks indicating an earlier form of the EBPB format with only the serial number following

I'm OK with not supporting disks formatted with MS-DOS before 4.0.

jannic (approved):

PR #147, opened 2024-10-11
Use chunks_exact and remove 'reason' from BlockDevice API

thejpster:
Some clean ups in the EuroRust 2024 impl room.

) Removed the reason field from the Block Device API. It didn't get used much and the caller can just do the logging. We didn't even log writes and they're more important.
) Addeded trace logging for reads/writes.
) Use chunks_exact to simplfy some loops


jannic (approved):

thejpster:
Needs a rebase.

PR #148, opened 2024-10-11
Add a universal block cache

thejpster:
Adds a block cache into the `VolumeManager` and redirect all block reads and writes though it. Avoids the need to ever stack allocate a 512 byte block (or two).

Based on #147.

### DIFF
--- a/src/blockdevice.rs
+++ b/src/blockdevice.rs
@@ -16,40 +16,6 @@ pub struct Block {
     pub contents: [u8; Block::LEN],
 }
 
-/// The linear numeric address of a block (or sector).
-///
-/// The first block on a disk gets `BlockIdx(0)` (which usually contains the
-/// Master Boot Record).
-#[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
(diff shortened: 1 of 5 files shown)


thejpster:
I realise now that this is completely incompatible with #143. I'll have to have a think about how to fix that.

jannic:
> I realise now that this is completely incompatible with #143. I'll have to have a think about how to fix that.

Would it be possible to move the `BlockCache` into `data :RefCell` ?

thejpster:
> > I realise now that this is completely incompatible with #143. I'll have to have a think about how to fix that.
> 
> Would it be possible to move the `BlockCache` into `data :RefCell` ?

Maybe ... but I return a reference to the `Block`.

https://doc.rust-lang.org/stable/std/cell/struct.RefMut.html#method.map_split might help.

thejpster:
Yeah, turns out its fine to put the cache in the RefCell with all the other data.

thejpster:
This one is good to go.

jannic (approved):

PR #151, opened 2024-10-12
Better trace logs.

thejpster:
A nice easy one. I was looking at trace logs from the `shell` program and "Reading FAT" was not useful.


eldruin (approved):
Looks good to me, thanks!

PR #152, opened 2024-10-16
Create a VolumeManager type alias

thejpster:
Makes the example code a bit simpler.


jannic (approved):
Agree, that looks better.

PR #153, opened 2024-10-17
Make read/write async

brunob45:


brunob45:
This was not intended

Issue #50, opened 2022-02-21
Support for asynchronous I/O


TilBlechschmidt:
I am currently experimenting with the nRF52 and using the [embassy](https://embassy.dev/embassy/dev/index.html) project. I intend to use an SD Card in the project. However, this library does not appear to support asynchronous APIs. This includes two levels:

1. DMA related SPI store/fetch operations
  - You give it a tx+rx buffer, it at some point interrupts you once everything has been sent/received
  - Allows the user/application to do other work in the meantime
  - Currently unsupported by the `SdMmcSpi` struct as it solely uses byte-by-byte synchronous send/receive methods
2. Asynchronous block devices
  - Most disk I/O is inherently asynchronous as the CPU waits for the device
  - In a majority of use cases microcontrollers might not have much else to do
  - In some cases, other, time-critical tasks need to run (e.g. handling wireless transmissions)

---

**My question here is:** _Would you be interested in integrating support for asynchronous SPI backends?_
**Follow-up question:** _If yes, do you have any thoughts on how to do that besides just "implementing everything twice"_

I can definitely see both arguments; for integrating it to have a unified place to look for when using SD cards; against stuffing everything into one crate even though it is basically its own thing (sadly, we can't simply pick and choose between async/sync APIs like we can with traits/generics).

ninjasource:
Hi, I've been experimenting with implementing this and have the following to report. 

- First, the good news. I got async file reads working and it's running about 8x faster than the blocking version on my nrf52480 chip using Embassy and DMA. The added benefit being that it does not hold up the mcu which was the reason why I needed the functionality in the first place (I'm working with audio).
- Now the bad news. This pretty much duplicates almost all of the codebase to the point where putting this in a separate repo or separate member of the same repo makes more sense to me than polluting this repo with feature flags in an effort to reuse some of the code. As mentioned previously, async is indeed very pervasive and duplicating everything like this is distasteful. I can't really see a way around this unfortunately.
- Dependency wise, the change essentially swaps `embedded-hal` with `embedded-hal-async` which is alpha right now but it is a member of the `embedded-hal` repo so it should be standard. This means that a nightly toolchain is needed with `#![feature(async_fn_in_trait, impl_trait_projections)]` features enabled. 
[...]

PR #154, opened 2024-10-18
Add maybe_async

brunob45:
I'm using embassy, and the other repo I found (embedded-fatfs) is not supporting partition tables on the SD card.
Here is my proposal to add `maybe_async` to this repo, so it can be used "as-is", or asynchronously.

I also needed to make `BlockDevice` mutable to support DMA on STM32.

All comments are welcomed.

### DIFF
--- a/src/filesystem/files.rs
+++ b/src/filesystem/files.rs
@@ -3,7 +3,12 @@ use crate::{
     filesystem::{ClusterId, DirEntry, Handle},
     BlockDevice, Error, RawVolume, VolumeManager,
 };
+
+#[cfg(feature = "is_sync")]
 use embedded_io::{ErrorType, Read, Seek, SeekFrom, Write};
+
+#[cfg(not(feature = "is_sync"))]
+use embedded_io_async::{ErrorType, Read, Seek, SeekFrom, Write};
 
 /// A handle for an open file on disk.
 ///
@@ -76,13 +81,15 @@ where
     /// Read from the file
     ///
     /// Returns how many bytes were read, or an error.
-    pub fn read(&self, buffer: &mut [u8]) -> Result<usize, crate::Error<D::Error>> {
-        self.volume_mgr.read(self.raw_file, buffer)
+    #[maybe_async::maybe_async]
+    pub async fn read(&self, buffer: &mut [u8]) -> Result<usize, crate::Error<D::Error>> {
+        self.volume_mgr.read(self.raw_file, buffer).await
     }
 
     /// Write to the file
-    pub fn write(&self, buffer: &[u8]) -> Result<(), crate::Error<D::Error>> {
-        self.volume_mgr.write(self.raw_file, buffer)
+    #[maybe_async::maybe_async]
+    pub async fn write(&self, buffer: &[u8]) -> Result<(), crate::Error<D::Error>> {
+        self.volume_mgr.write(self.raw_file, buffer).await
     }
 
     /// Check if a file is at End Of File.
@@ -130,16 +137,18 @@ where
     }
 
     /// Flush any written data by updating the directory entry.
-    pub fn flush(&self) -> Result<(), Error<D::Error>> {
-        self.volume_mgr.flush_file(self.raw_file)
+    #[maybe_async::maybe_async]
+    pub async fn flush(&self) -> Result<(), Error<D::Error>> {
+        self.volume_mgr.flush_file(self.raw_file).await
     }
 
     /// Consume the `File` handle and close it. The behavior of this is similar
     /// to using [`core::mem::drop`] or letting the `File` go out of scope,
     /// except this lets the user handle any errors that may occur in the process,
     /// whereas when using drop, any errors will be discarded silently.
-    pub fn close(self) -> Result<(), Error<D::Error>> {
-        let result = self.volume_mgr.close_file(self.raw_file);
+    #[maybe_async::maybe_async]
+    pub async fn close(self) -> Result<(), Error<D::Error>> {
+        let result = self.volume_mgr.close_file(self.raw_file).await;
         core::mem::forget(self);
         result
     }
@@ -186,11 +195,12 @@ impl<
         const MAX_VOLUMES: usize,
     > Read for File<'_, D, T, MAX_DIRS, MAX_FILES, MAX_VOLUMES>
 {
-    fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
+    #[maybe_async::maybe_async]
+    async fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
         if buf.is_empty() {
             Ok(0)
         } else {
-            self.read(buf)
+            Self::read(self, buf).await
         }
     }
 }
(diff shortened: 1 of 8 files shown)


brunob45:
Related to #50

src/lib.rs
```diff
@@ -301,7 +307,7 @@ where
     T: crate::TimeSource,
 {
     raw_volume: RawVolume,
-    volume_mgr: &'a VolumeManager<D, T, MAX_DIRS, MAX_FILES, MAX_VOLUMES>,
+    volume_mgr: &'a mut VolumeManager<D, T, MAX_DIRS, MAX_FILES, MAX_VOLUMES>,
```
thejpster, line 310: I think this breaks the ability to open multiple volumes/directories/files at a time, which I added recently?

thejpster:
It might be worth looking at #148 where I put a `BlockCache` in a refcell. That might help you have a mutable block device without making all the `&self` methods back to `&mut self` (which forces you to use raw handles if you want to open multiple files at once).

thejpster:
So for this to fly, I would need the examples to build-as is, so that we know we're not breaking existing code (at least more than I have already broken it). But perhaps also an async example would be useful. And it has to build on stable.

src/lib.rs
brunob45: Oh, didn't know that.. I guess the borrow checker would not like that

brunob45:
That's fair, I'll work on that

Otherwise, I know there is a lot of debates around `maybe_async`, and I'd understand if this PR is not the way forward for this library. What's your thoughts on this?

thejpster:
I'll judge it by the outcome. If the examples build unchanged on stable (and it works on no_std)  then it should continue to work for my use case and I can support the changes.

thejpster:
#148 was merged, so this will need a rebase.

src/filesystem/files.rs
```diff
@@ -225,7 +237,8 @@ impl<
         const MAX_VOLUMES: usize,
     > Seek for File<'_, D, T, MAX_DIRS, MAX_FILES, MAX_VOLUMES>
 {
-    fn seek(&mut self, pos: SeekFrom) -> Result<u64, Self::Error> {
+    #[maybe_async::maybe_async]
```
brunob45, line 240: must be async because it's part of `embedded-io-async`

brunob45:
@thejpster all fixed!

brunob45:
Run workflow?

thejpster:
We'll want to find a way to do a build in both async mode and non-async-mode. I think you'll need to mark the existing examples as requiring the `is_sync` feature, and make a new async example which only builds when that isn't set? Actually I don't even know if that's possible. The async example might have to be in a crate all of its own.

thejpster:
> Run workflow?

As this PR is not from an org member, one of us has to manually approve the workflow run every time - in case you submit a change which copies a Github secret key to pastebin or something. Sorry about that - bear with us. Alternatively you can look at the workflow files and get a good idea about what you need to run locally to replicate the CI.

thejpster:
> formatting is creating lots of diffs

We just test for `cargo fmt --check` so you probably just need to run `cargo fmt` locally.

thejpster:
> Expected — Waiting for status to be reported 

unfortunately it looks like adding the features to the matrix has renamed the build jobs, and so now the branch protection rule is waiting for a job that doesn't exist. I guess I'll have to manually override that check when this is ready to go, and then update the branch protection rule.

GHA is such a pain to use.

brunob45:
Thanks for the information! For a moment I thought the workflows were triggered by commenting here...

I'll try to put the is_sync feature outside the matrix, that should resolve the renaming of workflow

I just pushed the result of `cargo fmt`

brunob45:
Oh, I ran cargo fmt on windows... Is it different than Linux?

thejpster:
Try using 1.81. Maybe GitHub hasn't updated the image yet.

brunob45:
I had a rustfmt.toml file in the parent directory, limiting the line length to 120. I removed it and it fixed the issue

ninjasource:
> We'll want to find a way to do a build in both async mode and non-async-mode. I think you'll need to mark the existing examples as requiring the `is_sync` feature, and make a new async example which only builds when that isn't set? Actually I don't even know if that's possible. The async example might have to be in a crate all of its own.

IMO, this is the biggest issue with using the `maybe_async` crate. It breaks cargo's rules about features being additive. I don't think that it would be inconceivable for a user to need both blocking and non-blocking or one crate to use blocking (say for logging) and another non-blocking. Ideally the macro would generate separately named functions so that crates can play nicely with each other but alas it doesn't.

ivmarkov:
> > We'll want to find a way to do a build in both async mode and non-async-mode. I think you'll need to mark the existing examples as requiring the `is_sync` feature, and make a new async example which only builds when that isn't set? Actually I don't even know if that's possible. The async example might have to be in a crate all of its own.
> 
> IMO, this is the biggest issue with using the `maybe_async` crate. It breaks cargo's rules about features being additive. I don't think that it would be inconceivable for a user to need both blocking and non-blocking or one crate to use blocking (say for log flushing which may have a blocking api) and another non-blocking. Ideally the macro would generate separately named functions so that crates can play nicely with each other but alas it doesn't.

I've found a workaround in simply not using the `is-sync` feature at all, and then also using the `duplicate` crate which allows you to re-export a single, templatized module twice:
- once, as blocking
- one more time, as async

... so you are effectively ending up _simultaneously_ with two versions of the code, each in its submodule. Modulo the two submodules which of course have different names, the lib otherwise is having a mostly identical api for blocking and async.

A bit like having `embedded-hal` and `embedded-hal-async` in the same crate under two different modules.

Note that I've only tried this with smaller drivers where I could fit all "maybe-async" code in a single module. Not sure how it would work for larger codebase split across many modules.

ivmarkov:
Here's the relevant comment in the maybe-async crate that I follow myself: https://github.com/fMeow/maybe-async-rs/issues/6#issuecomment-1893918591

elpiel:
Thank you very much for working on this @brunob45 .
I wonder if you can also tweak the SdCard impl as I personally want to use the raw read/write support  instead of the fat32 one.

ValouBambou:
Is this still planned to be merged? If so, I'm interested in helping if needed. 

Also, I noticed that in the implementation of  `SdCard`  there are calls to `embedded_hal::delay::DelayNs::delay_us`. This is fine for blocking, but maybe for async version we could instead use the equivalent method from `embedded-hal-async` and enable a sleep with async capabilities. There are maybe other places where we can do better than just forwarding the async/await keywords, like with the `maybe_async` macro.

thejpster:
I think it still needs the SD card driver changing? And I'm still not entirely persuaded on the merits of this approach as compared to just making a different crate dedicated to async usage. Or just add partition support to that other crate.

Happy to take input from other maintainers.

Be-ing:
I want to see this crate support async, but I don't think maybe_async is the way to go. That crate has two huge flaws. The code the macro generates differs based on Cargo features _on the macro crate_, rather than generating feature-gated code compiled in the crate that the macro is used in. As @ninjasource noted, because Cargo features are additive, separate crates in the dependency graph using maybe_async [can break each other](https://github.com/fMeow/maybe-async-rs/issues/21). The other issue is that a crate using maybe_async can't be built simultaneously for synchronous and asynchronous usage; they're mutually exclusive because maybe_async uses the presence/absence of a single Cargo feature rather than separate Cargo features for sync and async.

The [bisync](https://docs.rs/bisync/latest/bisync/) crate looks like a much better alternative to me. I have opened a new PR that uses that instead of maybe_async: #176

> I'm still not entirely persuaded on the merits of this approach as compared to just making a different crate dedicated to async usage.

IMO this crate is way too big for putting sync and async versions in separate crates to be maintainable.

brunob45:
Thank you all for the comments and interest. I'll close this PR for now, as I'll be using `embedded-fatfs` for my async projects.

PR #155, opened 2024-10-20
Add LFN support

beregond:


thejpster:
Thank you for this pull request!

I have some comments.

* The nested array that was added uses 416 bytes of stack, on top of the heapless String of length 255. Passing the headless::String in would allow the caller to decide how much space they are prepared to use, plus UTF-8 is a more efficient encoding than an array of `char` and we can avoid the duplication.
* I didn't see where this checks the LFN hash and sequence number to check the LFN is valid.
* We probably also need to invalidate LFN entries when we delete a file.
* LFNs are equally valid on FAT16 and we should handle both. 
* We're going to need some tests ;)

I like the approach of doing the dir listing first. That makes sense.

beregond:
Yes, I will provide code for fat16 as well as all the tests - for now it is a little mess (as I'm running this against actual SD card to understand fully what I'm trying to implement), but will do :+1:

jonathanpallant:
You inspired me to have a go at this myself - see https://github.com/rust-embedded-community/embedded-sdmmc-rs/tree/add-lfn-support

beregond:
Nice! Therefore I will continue to work on this branch until I loose interest - I treat this as educational thing as well (and definitely will take some ideas from your branch, I see solutions for problems that I've been stuck on a bit)

thejpster:
Please do feel free to work on this. Perhaps there's a better API than the one we have. But I'll close this PR in the meantime.

PR #157, opened 2024-10-25
Add LFN support when iterating directories

jonathanpallant:
You can now optionally see the long file names when doing directory listings. To do so you must pass an `LfnBuffer` of your chosen size to the new `iterate_dir_lfn` methods.


thejpster:
ugh, dealing with the fact that one unicode character might become two UTF-16 code units, and those two code units could be split across two directory entries, was a nightmare. But I think this works.

eldruin (commented):
Looks good to me, thanks!
Could you add an entry to the changelog about this?

thejpster:
> Looks good to me, thanks! Could you add an entry to the changelog about this?

Done

eldruin (approved):
Thank you for your work!

PR #158, opened 2024-10-29
Move FAT-specific code in make_dir_in_dir (and fix .. entries)

### LINKED ISSUE #161: When parent (`..`) of dir is root, use cluster ID 0
Otherwise you get errors from fsck like:

```text
/DIR3
  Invalid '..' entry in the second slot. Fixing.
```

thejpster:
This code belongs in the FAT implementation as it is all FAT-specific.

Also fixed a bug where `..` had the wrong parent cluster - both FAT16 and FAT32 should use 0 according to the complaints I was getting from fsck.

Fixes #161


thejpster:
Rebased on develop

src/volume_mgr.rs
```diff
@@ -1072,77 +1072,13 @@ where
             VolumeType::Fat(fat) => {
                 // TODO: Move this into the FAT volume code
```
jannic, line 1073: The TODO is resolved by this merge request and should be removed.

jannic (approved):

PR #164, opened 2024-10-30
Backport some fixes to 0.8 branch

thejpster:
* Write out info sector when volume closed
* Get `..` dir entry correct
* Write to second FAT

Now fsck should stop complaining about volumes that were written to with this crate.


thejpster:
I'll publish this as 0.8.1

jannic (approved):
An update of CHANGELOG.md is missing, but that's true for #160 as well, and it looks like in this repo the changelog update is often done manually before a release. So it's fine for now - please just don't forget it before releasing 0.8.1.

PR #167, opened 2024-11-22
Set rust-version to 1.76 & add matrix for building and testing for MSRV

### LINKED ISSUE #166: Consider an appropriate MSRV
Since we are moving forward with implementations that will bring async support I think it's good time to talk about MSRV.

`async_fn_in_trait` has been stabalised in `1.75.0` and I think we can start from there.
I can:
- [ ] tweak the CI to include MSRV build
as well as
- [ ] change the Cargo.toml to include `rust-version = "1.75"`

elpiel:
closes #166

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -6,7 +6,10 @@ jobs:
   formatting:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v4
+    - uses: dtolnay/rust-toolchain@stable
+      with:
+        components: rustfmt
     - name: Check formatting
       run: cargo fmt -- --check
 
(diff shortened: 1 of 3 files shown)


thejpster:
> exclusive range pattern syntax is experimental

oops. Should be easily fixable.

elpiel:
Yeah, they are easy to fix.
Will get to that.

elpiel:
I fixed the pattern matching but this `trim_ascii_end` is giving me a headache.
Any ideas how to replace it?

thejpster:
You could just copy/paste the source code from the newer library:

```rust
impl VolumeName {
    pub fn name(&self) -> &[u8] {
        let mut bytes = &self.0[..];
        while let [rest @ .., last] = bytes {
            if last.is_ascii_whitespace() {
                bytes = rest;
            } else {
                break;
            }
        }
        bytes
    }
}
```

elpiel:
Yep, that should do the trick.
I think I initially just used `&self.content` which didn't work.

Checks are passing now :tada:

thejpster:
Yeah because Volume Names are padded with space characters.

thejpster:
The build tasks have been renamed so CI is sad because the job it wants to see didn't run.

The configuration for which jobs are mandatory should be in git, but it's not. It's hidden in GitHub's proprietary back-end.

PR #170, opened 2024-12-22
Introduce a generic `Transport` trait to support multiple SD card protocols

luojia65:
Currently, the `embedded-sdmmc` library only supports SPI-based SD card access. However, certain MCUs and SoC chips feature dedicated SD protocol peripherals that can provide faster and more efficient access. This PR abstracts various SD card protocols into a new `Transport` trait, enabling future extensions and custom transports by third-party libraries.

```rust
/// Abstract SD card transportation interface.
pub trait Transport {
    /// Read one or more blocks, starting at the given block index.
    fn read(&mut self, blocks: &mut [Block], start_block_idx: BlockIdx) -> Result<(), Error>;
    /// Write one or more blocks, starting at the given block index.
    fn write(&mut self, blocks: &[Block], start_block_idx: BlockIdx) -> Result<(), Error>;
    /// Determine how many blocks this device can hold.
    fn num_blocks(&mut self) -> Result<BlockCount, Error>;
    /// Return the usable size of this SD card in bytes.
    fn num_bytes(&mut self) -> Result<u64, Error>;
    /// Can this card erase single blocks?
    fn erase_single_block_enabled(&mut self) -> Result<bool, Error>;
    /// Check the card is initialised.
    fn check_init(&mut self) -> Result<(), Error>;
    /// Mark the card as requiring a reset.
    ///
    /// The next operation will assume the card has been freshly inserted.
    fn mark_card_uninit(&mut self);
    /// Get the card type.
    fn get_card_type(&self) -> Option<CardType>;
    /// Tell the driver the card has been initialised.
    unsafe fn mark_card_as_init(&mut self, card_type: CardType);
}
```

This pull request includes:
- Refactor SPI-based SD card logic: moves existing SPI-specific functionality into a dedicated module, introduces SpiTransport as the default SPI transport implementation.
- Introduce the Transport trait: defines a consistent interface for reading, writing, and managing SD card blocks, makes the codebase more flexible for additional SD protocols or hardware-accelerated peripherals.
- SdCard wrapper changes: updates `SdCard` to hold a mutable reference to a `Transport` implementor, renames constructors to `new_spi` and `new_spi_with_options` (breaking) to clarify SPI usage.

These changes pave the way for supporting more advanced SD interfaces in hardware, improving performance and enabling community-driven protocol expansions.

r? @thejpster

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog] and this project adheres to [Semantic 
 - __Breaking Change__: The `VolumeManager::device` method now takes a callback rather than giving you a reference to the underlying `BlockDevice`
 - __Breaking Change__: `Error:LockError` variant added.
 - __Breaking Change__: `SearchId` was renamed to `Handle`
+- __Breaking Change__: sdcard!: add `Transport` trait, make `SdCard` an abstract of `Transport` other than `SpiDevice<u8>`, refactor internal implementation, add `new_spi` and `new_spi_with_options` functions.
 
 ### Added
 
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ You will need something that implements the `BlockDevice` trait, which can read 
 
 ```rust
 // Build an SD Card interface out of an SPI device, a chip-select pin and the delay object
-let sdcard = embedded_sdmmc::SdCard::new(sdmmc_spi, delay);
+let sdcard = embedded_sdmmc::SdCard::new_spi(sdmmc_spi, delay);
 // Get the card size (this also triggers card initialisation because it's not been done yet)
 println!("Card size is {} bytes", sdcard.num_bytes()?);
 // Now let's look for volumes (also known as partitions) on our block device.
--- a/examples/readme_test.rs
+++ b/examples/readme_test.rs
@@ -120,7 +120,7 @@ fn main() -> Result<(), Error> {
     // END Fake stuff that will be replaced with real peripherals
 
     // Build an SD Card interface out of an SPI device, a chip-select pin and the delay object
-    let sdcard = embedded_sdmmc::SdCard::new(sdmmc_spi, delay);
+    let sdcard = embedded_sdmmc::SdCard::new_spi(sdmmc_spi, delay);
     // Get the card size (this also triggers card initialisation because it's not been done yet)
     println!("Card size is {} bytes", sdcard.num_bytes()?);
     // Now let's look for volumes (also known as partitions) on our block device.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,7 +25,7 @@
 //!     D: embedded_hal::delay::DelayNs,
 //!     T: TimeSource,
 //! {
-//!     let sdcard = SdCard::new(spi, delay);
+//!     let sdcard = SdCard::new_spi(spi, delay);
 //!     println!("Card size is {} bytes", sdcard.num_bytes()?);
 //!     let volume_mgr = VolumeManager::new(sdcard, ts);
 //!     let volume0 = volume_mgr.open_volume(VolumeIdx(0))?;
--- a/src/sdcard/mod.rs
+++ b/src/sdcard/mod.rs
@@ -51,8 +51,8 @@ where
     /// deferred until a method is called on the object.
     ///
     /// Uses the default options.
-    pub fn new(spi: SPI, delayer: DELAYER) -> Self {
-        Self::new_with_options(spi, delayer, AcquireOpts::default())
+    pub fn new_spi(spi: SPI, delayer: DELAYER) -> Self {
+        Self::new_spi_with_options(spi, delayer, AcquireOpts::default())
     }
 
     /// Construct a new SD/MMC Card driver, using a raw SPI interface and the given options.
@@ -62,7 +62,7 @@ where
     ///
     /// The card will not be initialised at this time. Initialisation is
     /// deferred until a method is called on the object.
-    pub fn new_with_options(spi: SPI, delayer: DELAYER, options: AcquireOpts) -> Self {
+    pub fn new_spi_with_options(spi: SPI, delayer: DELAYER, options: AcquireOpts) -> Self {
         SdCard {
             inner: RefCell::new(spi::SpiTransport::new(spi, delayer, options)),
         }
--- a/src/sdcard/mod.rs
+++ b/src/sdcard/mod.rs
@@ -5,6 +5,7 @@
 
 pub mod proto;
 mod spi;
+pub use spi::SpiTransport;
 
 use crate::{Block, BlockCount, BlockDevice, BlockIdx};
 use core::cell::RefCell;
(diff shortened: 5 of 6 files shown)


thejpster:
This looks really interesting - thank you for the PR!

Did you have a specific SD Host Controller you wanted to implement `Transport` for?

thejpster:
Currently most of the user-facing types inside `mod sdcard` are re-exported at the top-level to make them easier to find. Should we also re-export `sdcard::Transport` and `sdcard::SpiTransport`?

thejpster:
Reading through this, I think a bunch of the implementation detail in the `SpiTransport` struct could be lifted out into the `SdCard` struct, and the `Transport` trait changed to offer:

* `fn card_command(&mut self, command: u8, arg: u32) -> Result<u8, Error>;`
* `fn card_acmd(&mut self, command: u8, arg: u32) -> Result<u8, Error>;`
* `fn read_data(&mut self, buffer: &mut [u8]) -> Result<(), Error>;`
* `fn write_data(&mut self, buffer: &[u8]) -> Result<(), Error>;`

For example, I believe this method would be exactly the same regardless of which SD Card transport was used:

```rust
/// Read the 'card specific data' block.
fn read_csd(&mut self) -> Result<Csd, Error> {
    match self.card_type {
        Some(CardType::SD1) => {
            let mut csd = CsdV1::new();
            if self.card_command(CMD9, 0)? != 0 {
                return Err(Error::RegisterReadError);
            }
            self.read_data(&mut csd.data)?;
            Ok(Csd::V1(csd))
        }
        Some(CardType::SD2 | CardType::SDHC) => {
            let mut csd = CsdV2::new();
            if self.card_command(CMD9, 0)? != 0 {
                return Err(Error::RegisterReadError);
            }
            self.read_data(&mut csd.data)?;
            Ok(Csd::V2(csd))
        }
        None => Err(Error::CardNotFound),
    }
}
```

But, I'd probably want to implement the `Transport` trait for something like the STM32 SD Host Controller before I'd know for sure, and I don't know that there's enough information in the Simplified SD Card Specification to do that, or if you need to see the full version.#

thejpster:
I pushed a few comment fixes to better track where things now are.

luojia65:
> This looks really interesting - thank you for the PR!
> 
> Did you have a specific SD Host Controller you wanted to implement `Transport` for?

@thejpster: Thanks for your feedback! I'm currently working on the following:
- SDH/SDIO peripheral for Bouffalo Lab chips (HAL support), which follows the standard SDHC register map.
- SMHC peripheral for Allwinner chips (HAL crates), which uses a non-standard SDHC register layout.

Several community members are also interested in adding SD protocol support for chips like STM32, ESP32, and others.

> Currently most of the user-facing types inside `mod sdcard` are re-exported at the top-level to make them easier to find. Should we also re-export `sdcard::Transport` and `sdcard::SpiTransport`?

Yes, I'd love to! If it's appropriate, I'll push a commit for this change.

> Reading through this, I think a bunch of the implementation detail in the `SpiTransport` struct could be lifted out into the `SdCard` struct, and the `Transport` trait changed to offer:
> 
>     * `fn card_command(&mut self, command: u8, arg: u32) -> Result<u8, Error>;`
>     * `fn card_acmd(&mut self, command: u8, arg: u32) -> Result<u8, Error>;`
>     * `fn read_data(&mut self, buffer: &mut [u8]) -> Result<(), Error>;`
>     * `fn write_data(&mut self, buffer: &[u8]) -> Result<(), Error>;`
> 
> 
> For example, I believe this method would be exactly the same regardless of which SD Card transport was used:
> 
> ```rust
> /// Read the 'card specific data' block.
> fn read_csd(&mut self) -> Result<Csd, Error> {
>     match self.card_type {
>         Some(CardType::SD1) => {
>             let mut csd = CsdV1::new();
>             if self.card_command(CMD9, 0)? != 0 {
>                 return Err(Error::RegisterReadError);
>             }
>             self.read_data(&mut csd.data)?;
>             Ok(Csd::V1(csd))
>         }
>         Some(CardType::SD2 | CardType::SDHC) => {
>             let mut csd = CsdV2::new();
>             if self.card_command(CMD9, 0)? != 0 {
>                 return Err(Error::RegisterReadError);
>             }
>             self.read_data(&mut csd.data)?;
>             Ok(Csd::V2(csd))
>         }
>         None => Err(Error::CardNotFound),
>     }
> }
> ```
> 
> But, I'd probably want to implement the `Transport` trait for something like the STM32 SD Host Controller before I'd know for sure, and I don't know that there's enough information in the Simplified SD Card Specification to do that, or if you need to see the full version.

I like the idea of reusing code from the SPI implementation. However, before proceeding, we should carefully review the SD card specification. There are notable differences between the SD and SPI protocols, such as variations in command reply types and initialization processes.

In fact, I attempted to provide a command-level abstraction in [this branch](https://github.com/luojia65/embedded-sdmmc-rs/tree/decouple-spi). However, I found that while it decouples SPI and SD protocols to some extent, it also makes it harder to create a unified abstraction for both.

Further exploration of the full SD Host Controller Specification might be necessary, especially for implementing the `Transport` trait for more complex systems like the STM32 SD Host Controller.

luojia65:
@thejpster: If we're unsure which functions to include in this trait, could we mark the entire trait as experimental or unstable for now? We can stabilize it later once we've finalized the functions to retain in `Transport`.

thejpster:
Are you planning on making Transport implementations outside this crate? I guess we could merge an experimental trait.

I see that embassy-stm32 has a working SD Host Controller driver. I looked at it briefly but I might dig into it a more. The Pico SDK also has an experimental driver that uses a custom PIO program.

luojia65:
> Are you planning on making Transport implementations outside this crate? I guess we could merge an experimental trait.
> 
> I see that embassy-stm32 has a working SD Host Controller driver. I looked at it briefly but I might dig into it a more. The Pico SDK also has an experimental driver that uses a custom PIO program.

luojia65:
> Are you planning on making Transport implementations outside this crate? I guess we could merge an experimental trait.
> 
> I see that embassy-stm32 has a working SD Host Controller driver. I looked at it briefly but I might dig into it a more. The Pico SDK also has an experimental driver that uses a custom PIO program.

@thejpster: Yes, I am planning to implement `Transport` in external crates. Having the trait available as an experimental feature would make it easier to prototype and test implementations for various platforms, such as those in embassy-stm32 or the Pico SDK. This approach would also allow us to refine the trait based on real-world use cases before finalizing its design.

thejpster:
Is a git dependency pointing at this branch sufficient?

luojia65:
> Is a git dependency pointing at this branch sufficient?

I’d suggest merging it into the upstream instead. While it would still require a Git dependency until a new version is published, merging into upstream ensures the changes are more visible and accessible to the community, encouraging broader testing and feedback. This can help refine the `Transport` trait more effectively.

thejpster:
OK. How do you want to mark the trait as 'unstable' then?

luojia65:
> OK. How do you want to mark the trait as 'unstable' then?

@thejpster: Can we introduce an `unstable` feature, similar to the `unproven` feature in embedded-hal 0.2? We could then mark the following items with `#[cfg(feature = "unstable")]`:

- `pub use crate::sdcard::{Transport, SpiTransport}` in lib.rs
- `pub use spi::SpiTransport` in sdcard.rs
- The `SdCard::new<T>(T)` constructor (not `SdCard::new_spi`)

This approach would allow experimental features to be gated while keeping the stable API intact.

thejpster:
Sounds OK to me. I guess the examples will tell us if we broke the API by doing this.

robamu:
@luojia65  Is there still activity/interest to get this merged? I started working on a SDIO driver on a Zynq7000 which also has a dedicated SDIO controller, so thihs looks very interesting.

nezza:
+1 on this one - it's a great improvement to the library and would help me embed my driver nicely.

sebgab:
+1 for me as well

thejpster:
I appreciate your interest in the library, but pinging me with comments like '+1' won't cause things to magically happen. I don't happen to need the proposed feature and so I won't be putting my own time into it, but if someone wants to update this PR, put it behind an 'unstable' feature flag and then implement it for a couple of different devices to show that the trait works in practice, that would be fantastic.

robamu:
I might do that soon. I have two potential boards with dedicated peripherals: An ARM9 based board with a MCI peripheral, and the Zynq7000 with a dedicated SDIO peripheral. I will see whether I can manage to make them work with embedded-sdmmc-rs first, and then look into integrating it with this PR.

PR #174, opened 2025-01-18
docs: README write to file example; reminder to run `file.flush()`

zpg6:
Adding example of writing to files in the code snippet. Includes a reminder to run `file.flush()` so that directory entry is updated (I has missed this resulting in #172).

```rust
let my_other_file = root_dir.open_file_in_dir("MY_DATA.CSV", embedded_sdmmc::Mode::ReadWriteCreateOrAppend)?;
my_other_file.write(b"Timestamp,Signal,Value\n")?;
my_other_file.write(b"2025-01-01T00:00:00Z,TEMP,25.0\n")?;
my_other_file.write(b"2025-01-01T00:00:01Z,TEMP,25.1\n")?;
my_other_file.write(b"2025-01-01T00:00:02Z,TEMP,25.2\n")?;

// Don't forget to flush the file so that the directory entry is updated
my_other_file.flush()?;
```

PR #175, opened 2025-01-18
docs: adds same snippet of writing to files to `lib.rs` for crate doc

zpg6:
Extension of #174 - adds same doc of snippet for writing to files to `lib.rs` so that it matches there. Tested running with `cargo doc` and it looks as expected.

image

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,15 +45,21 @@
 //!
 //! For writing files:
 //!
-//! ```rust,ignore
-//! let my_other_file = root_dir.open_file_in_dir("MY_DATA.CSV", embedded_sdmmc::Mode::ReadWriteCreateOrAppend)?;
-//! my_other_file.write(b"Timestamp,Signal,Value\n")?;
-//! my_other_file.write(b"2025-01-01T00:00:00Z,TEMP,25.0\n")?;
-//! my_other_file.write(b"2025-01-01T00:00:01Z,TEMP,25.1\n")?;
-//! my_other_file.write(b"2025-01-01T00:00:02Z,TEMP,25.2\n")?;
-//!
-//! // Don't forget to flush the file so that the directory entry is updated
-//! my_other_file.flush()?;
+//! ```rust
+//! use embedded_sdmmc::{BlockDevice, Directory, Error, Mode, TimeSource};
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -43,6 +43,19 @@
...
+//! ```rust,ignore
+//! let my_other_file = root_dir.open_file_in_dir("MY_DATA.CSV", embedded_sdmmc::Mode::ReadWriteCreateOrAppend)?;
+//! my_other_file.write(b"Timestamp,Signal,Value\n")?;
+//! my_other_file.write(b"2025-01-01T00:00:00Z,TEMP,25.0\n")?;
+//! my_other_file.write(b"2025-01-01T00:00:01Z,TEMP,25.1\n")?;
+//! my_other_file.write(b"2025-01-01T00:00:02Z,TEMP,25.2\n")?;
+//!
+//! // Don't forget to flush the file so that the directory entry is updated
+//! my_other_file.flush()?;
+//! ```
```
thejpster, line 57: If we wrap this code in a function, we can demonstrate that it does in fact compile.

```suggestion
//! ```rust
//! use embedded_sdmmc::{BlockDevice, Directory, Error, Mode, TimeSource};
//! fn write_file<D: BlockDevice, T: TimeSource, const DIRS: usize, const FILES: usize, const VOLUMES: usize>(
//!     root_dir: &mut Directory<D, T, DIRS, FILES, VOLUMES>,
//! ) -> Result<(), Error<D::Error>>
//! {
//!     let my_other_file = root_dir.open_file_in_dir("MY_DATA.CSV", Mode::ReadWriteCreateOrAppend)?;
//!     my_other_file.write(b"Timestamp,Signal,Value\n")?;
//!     my_other_file.write(b"2025-01-01T00:00:00Z,TEMP,25.0\n")?;
//!     my_other_file.write(b"2025-01-01T00:00:01Z,TEMP,25.1\n")?;
//!     my_other_file.write(b"2025-01-01T00:00:02Z,TEMP,25.2\n")?;
//!     // Don't forget to flush the file so that the directory entry is updated
//!     my_other_file.flush()?;
//!     Ok(())
//! }
//! ```
```

thejpster (approved):
Thanks!

PR #176, opened 2025-02-06
add async support using bisync crate

Be-ing:
This moves the existing API into a `blocking` module and duplicates the API into a new `asynchronous` module, using `async fn` where applicable. This uses the [bisync](https://docs.rs/bisync/latest/bisync/) crate rather than maybe_async used by #154. I think this resolves the concerns brought up in #154: sync and async can both be built at the same time, usage of bisync by other crates in the dependency graph doesn't affect this, and changes to the example and test code are minimal (just adjusting import path).

Currently there is some code that is identical between the sync and async implementations that is built twice. If you think using bisync is a direction worth pursing, I could move that to a common module if you want.

I have not added Cargo features for sync/async; they're both built. If requested, I could put each behind a feature gate, though I don't think it's necessary.

Fixes https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/50


thejpster (commented):
An interesting approach! I want to dig into the docs in more detail but it seems workable - modulo that AFIT feature-flag that has appeared.

thejpster:
It might also be good to move the example import renames to another PR to keep this one cleaner so we can see precisely only that which is affected by the async additions.

src/inner/mod.rs
```diff
@@ -0,0 +1,356 @@
+// ****************************************************************************
+//
+// Imports
+//
+// ****************************************************************************
+
+#![deny(missing_docs)]
+#![allow(async_fn_in_trait)]
```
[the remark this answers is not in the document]
Be-ing: No, it just silences a compiler warning that doesn't really matter for embedded. I suggest reading the discussion [on another embedded crate where I used this recently](https://github.com/smart-leds-rs/smart-leds-trait/pull/15) for background.

Be-ing:
> It might also be good to move the example import renames to another PR to keep this one cleaner so we can see precisely only that which is affected by the async additions.

That goes towards a bigger question: do you want to break the public API by moving the existing API into a new `blocking` module? This can be avoided by making the `blocking` module private and instead putting `pub use blocking::*;` in src/lib.rs. I don't think that would be great for downstream users in the long term though. My primary concern is that it would make the documentation confusing to navigate. Here's how the front page of the documentation would look with that:

image

There's the whole blocking API of the crate, then the `asynchronous` module... and when you go to the documentation for that, you see the same API:

image

I think this would be rather confusing. It's difficult to even tell that I clicked a link and navigated to another page because it looks so similar to the documentation's front page.

By moving the existing API into a `blocking` module (how this branch currently is), the front page of the documentation looks like this:

image

Be-ing:
I've avoided running `cargo fmt` for now to minimize the diff because it's rather big already. If you'd like that to be run now, just let me know, or I could do it later after you've reviewed.

src/inner/mod.rs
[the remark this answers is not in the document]
Be-ing: I added a comment explaining why this is there ([67c0134](https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/176/commits/67c0134af86637c5d357c4ce47358b2224797a72)). `async fn` in traits has been supported on stable Rust since 1.75 and the current MSRV for this crate is 1.76.

thejpster:
> I presume bad things would happen if those aren't closed manually?

Yes, leaking a file handle means you can't reopen it, and handles are a finite resource. Plus certain context is only written to disk on file close so the disk ends up needing a fsck/scandisk pass.

As long as that only affects the async users that's fine.

Moving the blocking API into the blocking module is fine with me.

thejpster:
Can you rebase? #177 might make this smaller.

thejpster:
Is it possible to give the async version a blocking drop? Better than doing nothing.

Be-ing:
> Can you rebase? https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/177 might make this smaller.

Okay, rebased.

> Is it possible to give the async version a blocking drop? Better than doing nothing.

Unfortunately not. That would require this crate to bring in its own async executor and set that up just within `impl Drop`. It's questionable if that would be desirable if it was even possible. In a library that uses std, you can wrap async functions with a simple async executor like [futures_lite::future::block_on](https://docs.rs/futures-lite/latest/futures_lite/future/fn.block_on.html) or [pollster::block_on](https://docs.rs/pollster/latest/pollster/fn.block_on.html), but both of those use std modules like std::future and std::sync that aren't available in core.

thejpster:
Is there no no_std block_on! ? Can't you just spin polling the Future?

Be-ing:
I just found [embassy_futures::block_on](https://docs.rs/embassy-futures/latest/embassy_futures/fn.block_on.html), however, I don't think it's a good idea to use it to impl Drop for async Files/Volumes. We can't know if a user would want that and there's no way to opt out of Drop::drop getting called if it's implemented. For example, the user could have strict power consumption requirements that require putting the MCU to sleep on every `.await` as Embassy and RTIC do.

thejpster:
But it's probably better than a corrupt file system, and any correct program will close them asynchronously before dropping.

What do other async File APIs do?

Be-ing:
> Suggestion: impl Drop for async File/Volume using embassy_futures::block_on, but put that behind an async_drop Cargo feature that is enabled by default. This will Just Work by default, and any project where that little bit of blocking is really a problem has an escape hatch.

Nevermind, I forgot `core::mem::forget` exists to opt out of Drop, and of course File/Volume::close already use it. There's no need for a Cargo feature to prevent dropping; that can be opted out of just by calling File/Volume::close.

Be-ing:
Some blogs about async drop:
https://without.boats/blog/asynchronous-clean-up/
https://sabrinajewson.org/blog/async-drop
https://smallcultfollowing.com/babysteps/blog/2023/03/16/must-move-types/

TL;DR: It's complicated and Rust is a long way from async drop being implemented. Probably not for some more years.

Blocking in Drop is an ugly workaround, but I think it's sufficient for now.

thejpster:
You'll need to handle dropping Directories too.

Be-ing:
Directory::drop doesn't make any async calls, so there's nothing to do there.

thejpster:
Ah! Yes you're right.

thejpster:
Ok, I need to find time to test this!

thejpster:
Still on my list - maybe after Embedded World

thejpster:
OK, this looks fine. If you can run a `cargo fmt` and rebase, let's get it done.

avsaase:
I just ran into a panic while using this branch when `drop` is called on an open volume. I'm using the embassy executor.

When `drop` is called I get this panic message
```
0.052407 ERROR panicked at 'Found waker not created by the Embassy executor. `embassy_time::Timer` only works with the Embassy executor.'
└─ embassy_executor::raw::waker::task_from_waker @ /home/alexander/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/embassy-executor-0.7.0/src/raw/waker.rs:38  
0.052459 ERROR panicked at /home/alexander/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/defmt-0.3.10/src/lib.rs:380:5:
explicit panic
└─ panic_probe::print_defmt::print @ /home/alexander/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/panic-probe-0.3.2/src/lib.rs:104 
```

Calling `block_on` from within an embassy executor is allowed by itself, so there must be something within the code that's called from `drop` that causes this. Or it's a bug in the embassy executor. The panic message is not very helpful since this crate doesn't depend on `embassy-time`.

I didn't dig any deeper yet but I created this repo that reproduces the problem: https://github.com/avsaase/embedded-sdmmc-async-drop.

ValouBambou:
This is a known issue if I remember correctly, it's intended behavior and the code from the user main should use the cargo feature `generic-queue-x`. This is documented [in the embassy-time documentation](https://docs.embassy.dev/embassy-time/git/default/index.html#generic-queue). It is maybe useful to add a note for that on the README and docs, but I'm not sure if this library should provide additional feature flags or other stuff to manage that.

avsaase:
Okay, so what happens is that the async `DelayNs` implementation of `embassy_time::Delay` awaits a timer which causes a panic when run in `embassy_futures::block_on` if `generic-queue-x` is not enabled. It would definitely be good to document this somewhere but I wonder if it's possible to avoid the problem altogether. One possible solution is to require users of the async api to pass in an implementer of both `embedded_hal::delay::DelayNs` and `embedded_hal_async::delay::DelayNs` and use the blocking `DelayNs` in the `Drop` impl. I haven't checked yet how pervasive this change would be in this code base. I can have a try over the next few days but it would be nice to hear from @thejpster and @Be-ing if they think this is good approach before I spent half my weekend on it.

Be-ing:
Thanks for testing and digging into this @avsaase and @ValouBambou! I documented the requirement for enabling a `generic-queue-x` feature on the embassy-time crate when dropping a File or Volume from an Embassy executor.

> One possible solution is to require users of the async api to pass in an implementer of both embedded_hal::delay::DelayNs and embedded_hal_async::delay::DelayNs and use the blocking DelayNs in the Drop impl. I haven't checked yet how pervasive this change would be in this code base. I can have a try over the next few days but it would be nice to hear from @thejpster and @Be-ing if they think this is good approach before I spent half my weekend on it.

That's an interesting idea, but I suspect it would be cumbersome for users, but perhaps more importantly, cumbersome to maintain. It's difficult to say whether it would be worthwhile and how pervasive the code changes would need to be before it's implemented. I suspect it wouldn't be worth the trouble because the issue is avoidable by manually calling `close` instead of letting the object drop. Ironically, the panic could be helpful to test to ensure that the File/Volume is *not* dropped accidentally.

Phirks:
I'm trying to get this working with Embassy on STM32F3 with a micro-SD card breakout board, but I'm having trouble with the basics (num_bytes returns CardNotFound error).  Here's my example repo:
https://github.com/Phirks/embassy-stm32f3-sd-dfu-example

When I check the communication, I'm sending 80 clock pulses then repeatedly CMD0 and getting back 0xFE and 0x07 over and over until it times out and sends a CardNotFound error.  It looks like it's expecting a 0x01.  I've tried setting the retries to 50,000 and it never works.  I've tried a few different sd cards of 2 different brands.  When I pull the card I just get 0xFF forever, which makes sense.
SD card response

I believe everything is set up here according to the docs and was hoping to get some help.  I had heard that clock cycles starting before the CS pin is set and remaining on after would be helpful, but I don't see how that would be controllable by a user with this interface.

thejpster:
> but I don't see how that would be controllable by a user with this interface.

It's not. You have to drive the SPI bus manually to do this. Some cards let you get away without it but it's in the spec and some cards require it. We can't do it for you without violating embedded-hal API design and we got told off for doing that.

yanshay:
I wanted to update that I got this async working on an ESP32-S3 based board - WT32-SC01-Plus. I've been using the standard crate for quite some time but the async was really missing.
So thanks for the hard work getting this Async done!

Along the way I had some issue that I'd like to understand before I fully switch to using Async.

When I started working, the Blocking version of my code worked fine and the Async failed with the following log:

```
WARN - Got response: 52, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..

> this went on about 50 lines

And then I receive an error:

====================== PANIC ======================
panicked at src/main.rs:142:72:
called `Result::unwrap()` on an `Err` value: DeviceError(CardNotFound)

Backtrace:

0x4200a2c9
0x4200a2c9 - esp32s3_async_sdmmc::____embassy_main_task::{{closure}}
    at ??:??
0x4200621a
0x4200621a - embassy_executor::raw::TaskStorage<F>::poll
    at /Users/yanshay/.cargo/registry/src/index.crates.io-6f17d22bba15001f/embassy-executor-0.7.0/src/raw/mod.rs:214
0x4200615d
0x4200615d - embassy_executor::raw::SyncExecutor::poll::{{closure}}
    at /Users/yanshay/.cargo/registry/src/index.crates.io-6f17d22bba15001f/embassy-executor-0.7.0/src/raw/mod.rs:430
0x4200d363
0x4200d363 - esp32s3_async_sdmmc::__xtensa_lx_rt_main
    at /Users/yanshay/Trials/esp32/esp32s3-async-sdmmc/src/main.rs:65
0x42013dd6
0x42013dd6 - Reset
    at /Users/yanshay/.cargo/registry/src/index.crates.io-6f17d22bba15001f/xtensa-lx-rt-0.18.0/src/lib.rs:82
0x40378c48
0x40378c48 - ESP32Reset
    at /Users/yanshay/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-hal-0.23.1/src/soc/esp32s3/mod.rs:165 
```

Sometimes it did work, usually if I ran the blocking version before, and only after that the async version if I didn't power the device down and only flashed the async version code (so the sync initialization seem to have initialized the device properly for the async to work). At least so it seemed, I made many runs so things weren't consistent.

Here is an example of the log when the file read succeeded, and when successful the sequence is always showed the exact same logs:

```
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 70, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 0, trying again..
WARN - Got response: 55, trying again..
WARN - Got response: 8, trying again..
WARN - Got response: 3f, trying again..
INFO - File content: This is content of file a.txt
```

And when it worked, I later had an embassy panic regarding waker that's not an embassy-executor waker.
That was until I realized I need to call `close().await`. The panic was on the Volume drop.

Then I made some changes, I'm not sure what, but they weren't that significant for initialization, and I think the only significant change is that I noticed I need to do `close().await` on the objects so not get that embassy panic.

I think that's when things started working fine and the above logs also disappeared and things started to work consistently. But I only think that, because I did all kinds of minor changes to the code while trying to get it to work.

- So my question is - could it be that not closing all objects properly using `close().await` caused some inconsistent state with the SDCard that would appear also in the initialization of the next run if device wasn't not turned off / or even if it was turned off?
I'm just trying to figure out if there's some reliability issue under certain circumstances and want to understand what caused the many failed runs above along the way before I fully switch to using async mode.

Be-ing:
That's a pretty bad failure mode and makes me wonder if we should have the option to Drop File and Volume. We could panic in those Drop impls with a helpful error messages saying to use `close().await` instead.

yanshay:
> That's a pretty bad failure mode and makes me wonder if we should have the option to Drop File and Volume. We could panic in those Drop impls with a helpful error messages saying to use `close().await` instead. Maybe we could make that the default Drop impl and add a Cargo feature `embassy-time-generic-queue` to opt into the `block_on` impls of Drop?
> 
> 
> 
> Thanks for all the testing and feedback everyone, it's is very useful!

That would have definitely been helpful and save quite some time. I would have never thought of the close of I didn't go through this tread and initially missed the await, only the warning made me notice that.

Other than that, could missing that be the cause for the failures I saw about CardNotFound? The drops come later so I wonder if the failures were due to not closing properly on earlier executions but I think it took place also right after power up of the device, only it doesn't any longer.

thejpster:
> could missing the proper close be the cause for the failures I saw about CardNotFound

Unlikely. Closing a file just involves writing out a new directory entry with the updated file size. But the card doesn't care what you write to it - it just sees block reads and block writes.

Be-ing:
> Unlikely. Closing a file just involves writing out a new directory entry with the updated file size. 

What about the Volume? Could the missing closure of the Volume cause what @yanshay encountered?

thejpster:
Nope. Nothing is written to the disk when you close a volume. The open volume just caches some filesystem metadata to save us looking it up. You can verify this by looking at the code that is executed when you close a volume.

thejpster:
SD Card initialisation is messy, and I've only tested this crate (and only in sync mode) on a limited number of cards. My recommendation is to strip everything down to the bare minimum, and put a logic analyser on the bus so you can see what's going to and from the card. And try the hardware with a known good SD Card implementation like fatfs. If fatfs works and this crate doesn't, we can look at the traces to try and work out why.

I speak from ~25 years of experience when I say that complex issues are not usually solved by trying things at random. You have to go back to the most basic state you can until you hit something solid, and then build up from there. If nothing seems solid, go lower down.

yanshay:
> SD Card initialisation is messy, and I've only tested this crate (and only in sync mode) on a limited number of cards. My recommendation is to strip everything down to the bare minimum, and put a logic analyser on the bus so you can see what's going to and from the card. And try the hardware with a known good SD Card implementation like fatfs. If fatfs works and this crate doesn't, we can look at the traces to try and work out why.
> 
> I speak from ~25 years of experience when I say that complex issues are not usually solved by trying things at random. You have to go back to the most basic state you can until you hit something solid, and then build up from there. If nothing seems solid, go lower down.

The thing is that now it's working, not sure what changes I did got it working.

So are you basically saying that the SDCard and not the Board hosting the SDCard is meaningful?
So even if it works for me on a specific SDCard it doesn't guarantee that it will work for someone else using the same board if they place a different brand of SDCard?
It's one thing if it fails for me, I can troubleshoot using LA, but if it happens for someone else it's a whole different story.

thejpster:
> So even if it works for me on a specific SDCard it doesn't guarantee that it will work for someone else using the same board if they place a different brand of SDCard?

Definitely. Each SD Card is a tiny computer running its own firmware, and they all do things slightly differently.

Be-ing:
> So are you basically saying that the SDCard and not the Board hosting the SDCard is meaningful?

Yes. My motivation for working on this is rewrite the firmware for a little [music player gadget](https://cooltech.zone/tangara/) in Rust. At first I just wanted to do this for fun, but I've been [struggling](https://codeberg.org/cool-tech-zone/tangara-fw/pulls/295#issuecomment-3117948) with bizarre, extreme slowdowns in read speeds with the C++ firmware, but only with one particular SD card. After trying everything I can with that firmware, my only idea left before concluding some interaction between the board and my particular SD card is cursed is trying a completely different software stack.

avsaase:
I think I ran into another problem or limitation. My SD card shares the SPI bus with another device using [`embassy_embedded_hal::shared_bus::asynch::spi::SpiDevice`](https://docs.embassy.dev/embassy-embedded-hal/git/default/shared_bus/asynch/spi/struct.SpiDevice.html). When writing to an open file I sometimes get a `DeviceError(WriteError)`, specifically from here: https://github.com/Be-ing/embedded-sdmmc-rs/blob/e7eef2a4a241bbc3fcc1e101b06c98adcc8e17bd/src/sdcard/mod.rs#L247-L249. Since the write operation can be interrupted at any `await` point, I suspect the shared bus causes a delay between sending the `CMD13` command and receiving the response byte and that this causes some sort of timeout in the card. A retry of the write seems to always succeed. So far I haven't seen any partially completed writes but it feels sketchy to rely on this.

I haven't had the time to probe it with a logic analyzer yet this is all still speculation at this point. I put together [this repoducer](https://github.com/avsaase/embedded-sdmmc-async-drop/blob/write-error-repro/src/main.rs) that triggers the error. [The blocking version](https://github.com/avsaase/embedded-sdmmc-async-drop/blob/write-error-repro-blocking/src/main.rs) works fine.

Be-ing:
SdCardInner::card_command already has a retry loop with a delay... :thinking: 

```rust
        let mut delay = Delay::new_command();
        loop {
            let result = self.read_byte().await?;
            if (result & 0x80) == ERROR_OK {
                return Ok(result);
            }
            delay
                .delay(&mut self.delayer, Error::TimeoutCommand(command))
                .await?;
        }
```

avsaase:
Yes. I added some logs like this
```rust
if self.card_command(CMD13, 0).await? != 0x00 {
    warn!("CMD13 failed");
    return Err(Error::WriteError);
}
if self.read_byte().await? != 0x00 {
    warn!("Read byte oyther than 0x00");
    return Err(Error::WriteError);
}
```
and it's consistently the the first one.

avsaase:
I forgot to mention that the reproducer occasionally shows a `DeviceError(DiskFull)` error as well. I haven't been able to reliably reproduce that one but I figured it's good to know.

Be-ing:
Is it reproducible with multiple SD cards?

avsaase:
> Is it reproducible with multiple SD cards?

I'm seeing the same error with a SanDisk Extreme U3 32GB and an off-brand U1 16GB card. Presumably the SanDisk card is of decent quality.

thejpster:
Interesting. Yeah, if the task running the SD card can be stalled for an arbitrary amount of time, we'll need to make sure we're robust to timeouts and make sure we retry things appropriately. That hasn't been an issue for me at all, because I'm single-tasking and using blocking mode.

avsaase:
I found a solution to to the `WriteError` problem by changing the way commands are sent to the card. Instead of writing the command bytes in one operation and then doing multiple single byte reads until `(result & 0x80) == ERROR_OK`, we can do both the write and the read in a single transfer. After the transfer is complete we scan the buffer for the response bytes. For the async API this is probably a performance improvement because the DMA can do a single transfer which can be awaited. A small downside is this likely clocks out a few more `0xFF` bytes from the card than strictly necessary but I don't think that's a problem.

According to [this Stack Overflow answer](https://electronics.stackexchange.com/a/602106/256470) (which gave me this idea), cards always start sending their responses within 1-8 bytes after the command bytes are sent. This puts an upper bound on the number of bytes in the transfer (19 to be precise).

I have a commit [here](https://github.com/avsaase/embedded-sdmmc-rs/commit/8fc205e15ac8999afad7d6084b9bbbb783e26594) that demonstrates this approach. I confirmed that this solves the problem shown in the reproducer. The blocking version also still works fine. The way I changed the `card_command` return type is not super elegant but this was necessary because now all the response bytes are read in one go, so they need to be passed back to the caller in case it needs to use the bytes after the first. This can probably be made more strongly typed but I wanted to get some feedback on this before spending too much time on it. 

I can make a PR against this branch or you can cherry pick the commit.

thejpster:
That SO answer is fantastic.

I am worried about the advice to send a dummy 0xFF byte with CS high in order to clear the output register on buggy cards.

Perhaps we should own two SpiDevice handles - one for the card and one for 'not the card'. Then we can bring the dummy byte clocking at start up back in-house.

avsaase:
It's slightly annoying that the `SpiDevice` trait doesn't have an escape hatch to manually control the CS line for doing non-standard stuff like this.

> Perhaps we should own two SpiDevice handles

I guess that could work but would that require the user to use a physical pin for the the "not the card" CS? That might not always be practical.

> bring the dummy byte clocking at start up back in-house.

In `embedded-fatffs` they have a [ `sd_init`](https://github.com/MabezDev/embedded-fatfs/blob/428ad03cee555d9636f061d7bca7a62c4889cc63/sdspi/src/lib.rs#L73-L83) function that takes a mutable reference to a `SpiBus` and a CS pin. This function could be copied as is. Unfortunately this pattern can't be used to send the `0xFF` byte with CS high because creating the spi device takes ownership of the CS pin.

avsaase:
I gave the dummy spi device approach a try but I encountered two problems:
1. With the `embassy-rp` HAL, and I assume other embassy HALs as well, it's not possible to not use a physical pin for the dummy CS. I tried creating my own dummy pin but somewhere in the trait spaghetti you always hit a sealed trait. Maybe there's another way to solve this.
2. In the async version, even with a dummy spi device there's no guarantee that the `0xFF` byte is written directly after the card spi transfer is complete because there's an await point. Depending on timings, another device might use the bus before the `0xFF` is written, negating any benefit. I don't see an obvious solution to this.

avsaase:
Yeah, that function only serves as a convenient way to send the 74 clock cycles. It doesn't solve the problem that we need to do stuff outside of the `SpiDevice` contract on every operation. I think the only proper way to do this is to extend `embedded-hal{-async}` to support custom behavior like this but I'm not holding my breath for that to happen.

Btw, I'm not sure if I've ever seen the lack of `0xFF` byte cause a problem. Although, from the explanation in that SO answer it sounds like this could cause strange things to happen. [This page](http://elm-chan.org/docs/mmc/mmc_e.html#spibus) from the SdFat C library alsa says it's important.

Be-ing:
> I am worried about the advice to send a dummy 0xFF byte with CS high in order to clear the output register on buggy cards.

If this is working with multiple SPI devices on a bus without doing that, do we need to bother?

Be-ing:
> In embedded-fatffs they have a[ sd_init](https://github.com/MabezDev/embedded-fatfs/blob/428ad03cee555d9636f061d7bca7a62c4889cc63/sdspi/src/lib.rs#L73-L83) function that takes a mutable reference to a SpiBus and a CS pin. This function could be copied as is. Unfortunately this pattern can't be used to send the 0xFF byte with CS high because creating the spi device takes ownership of the CS pin.

A nice thing about this API is that it allows the caller to get back a mutable reference to the SPI device to change the clock speed after initialization.

src/inner/sdcard/mod.rs
```diff
@@ -382,182 +398,198 @@ where
...
+        buf[3] = (arg >> 8) as u8;
+        buf[4] = arg as u8;
         buf[5] = crc7(&buf[0..5]);
 
-        self.write_bytes(&buf)?;
-
-        // skip stuff byte for stop read
-        if command == CMD12 {
-            let _result = self.read_byte()?;
-        }
```
avsaase, line 506: I didn't add this check to the buffer scan. Do we know what the value of this byte is? I can't find anything about it in the spec.

Jmlannan: From my reading this is not an official spec but it is commonly discarded 
i.e. http://elm-chan.org/docs/mmc/mmc_e.html
"The received byte immediataly following CMD12 is a stuff byte, it should be discarded prior to receive the response of the CMD12"
Seems unlikely that it will contain 0x80 and it seems other implementations allow this byte to be check anyway like [micropython sdcard.py](https://github.com/micropython/micropython-lib/blob/master/micropython/drivers/storage/sdcard/sdcard.py).

Jmlannan: Actually sorry, that example does show that byte being skipped with if self.cmd(12, 0, 0xFF, skip1=True). Perhaps this is important for some cards?

Jmlannan: Im not sure that when issuing CMD12 it is checking for errors well (here or main).  If a transfer is not actually stopped then it may get stuck in block reading unexpectedly and not pass an error.  But checking for a valid response from CMD12 may not really be an issue here since it will just read a bunch of bytes that should get the card back to a r1 state anyway.

yanshay:
Hi, 
What's the status of this PR? Any plans to merge it? Any known reasons not to use this - I mean, is there any known experience that it's less good than the sync version?
I started using it and as long as I close() things and don't let them drop it seems to work fine.
The 0xFF issue mentioned above seems to be not specific to the async version, right?

marcfir:
> I found a solution to the `WriteError` problem by changing the way commands are sent to the card. Instead of writing the command bytes in one operation and then doing multiple single byte reads until `(result & 0x80) == ERROR_OK`, we can do both the write and the read in a single transfer. After the transfer is complete we scan the buffer for the response bytes. For the async API this is probably a performance improvement because the DMA can do a single transfer which can be awaited. A small downside is this likely clocks out a few more `0xFF` bytes from the card than strictly necessary but I don't think that's a problem.
> 
> According to [this Stack Overflow answer](https://electronics.stackexchange.com/a/602106/256470) (which gave me this idea), cards always start sending their responses within 1-8 bytes after the command bytes are sent. This puts an upper bound on the number of bytes in the transfer (19 to be precise).
> 
> I have a commit [here](https://github.com/avsaase/embedded-sdmmc-rs/commit/8fc205e15ac8999afad7d6084b9bbbb783e26594) that demonstrates this approach. I confirmed that this solves the problem shown in the reproducer. The blocking version also still works fine. The way I changed the `card_command` return type is not super elegant but this was necessary because now all the response bytes are read in one go, so they need to be passed back to the caller in case it needs to use the bytes after the first. This can probably be made more strongly typed but I wanted to get some feedback on this before spending too much time on it.
> 
> I can make a PR against this branch or feel free to cherry pick the commit.

The single transfer is good, except for CMD17 and CMD18, since this transfer can contain the `DATA_START_BLOCK`.
I provided a fix for this issue at https://github.com/Be-ing/embedded-sdmmc-rs/pull/2.

PegasisForever:
@avsaase I believe your https://github.com/Be-ing/embedded-sdmmc-rs/pull/1 breaks CMD9 for reading csd (and potentially other commands).

I'm testing with a SDHC card on stm32G4 with embassy framework.

This is the correct csd data the library is supposed to read after issuing CMD9: `[64, 14, 0, 50, 91, 89, 0, 0, 28, 223, 127, 128, 10, 64, 0, 215]`
After the pr, this is the command response returned from `card_command(CMD9, 0)`: `[0, 255, 254, 64, 14]`, and the `read_data(&mut csd.data)` following that throws an `ReadError` because the first byte it got is `223` instead of the expected `0xFE`.
You can tell the `card_command(CMD9, 0)` is eating up some data that is supposed to be received by `read_data`. (`..., 64, 14` part), also `223` is one of the numbers in the actual csd data.

avsaase:
Did you try @marcfir's linked PR? That seems to address the issues your describe.

The problem here is that if the task gets stalled after sending the command the card can hit a timeout, failing the transfer. At least that is my assumption. Ideally we'd have some way to not returning to the executor after sending the command while we wait for the right response bytes but as far as I know this kid of mixing of sync and async operations is not possible.

thejpster:
The whole point of having an async SPI transport is that you can do other things whilst the data is being sent over SPI (e.g. using DMA), and if you do an async operation in-between two SPI operations then yes, some other task has an opportunity to run.

It's definitely an argument for making an async-only SD card library - one that is cognizant of these issues and works around them.

PegasisForever:
> Did you try @marcfir's linked PR? That seems to address the issues your describe.
> 
> The problem here is that if the task gets stalled after sending the command the card can hit a timeout, failing the transfer. At least that is my assumption. Ideally we'd have some way to not returning to the executor after sending the command while we wait for the right response bytes but as far as I know this kid of mixing of sync and async operations is not possible.

No, that pr does not fix things, as that pr only changes behaviour for CMD17 and CMD18. I wonder how how he got CMD9 working on his cards, maybe its just card-to-card variance.

yanshay:
Any plans to merge this to the main release?

I'm using this for quite some time in a project and all in all its working great.

Only issue I've seen are two reports where a file ended up being length zero (code wise I couldn't see a flow that could cause this). Could it be due to the async nature or just a potential bug in the library unrelated to async?

thejpster:
As I understand it, there's still an issue where another async task might intervene during a card transaction and cause the card to timeout the command, which we don't currently handle. That would need to be fixed.

I still also need to decide of the pain of `bisync` is worth the benefit of 'async support' that I don't personally need or use right now. I'd certainly want some embedded async Rust experts on-board for maintenance, because I don't have any way of testing it.

thejpster:
There's also always https://github.com/MabezDev/embedded-fatfs/ as an alternative, which does appear to support async operation.

yanshay:
> As I understand it, there's still an issue where another async task might intervene during a card transaction and cause the card to timeout the command, which we don't currently handle. That would need to be fixed.

Could this be the reason for the zero length file? I definitely have many other async tasks, pretty complex application.

If after I write I test the file length and retry in case it's not as expected, will it use the cache for length information or will it check the sd card for real? If cache, any way to clear cache and test?

> 
> I still also need to decide of the pain of `bisync` is worth the benefit of 'async support' that I don't personally need or use right now. I'd certainly want some embedded async Rust experts on-board for maintenance, because I don't have any way of testing it.
>

First, it's working well so far for me and multiple users of my project. Other than two cases of zero length files after write.

And I don't see how any complex application can be written without async support, maybe there's some technique I'm not aware of but I run tens of tasks simultaneously and without async wouldn't be able to use this library. Just stating that from my POV it's definitely worth any complexity added.
Just my 2 cents.

zpg6:
> > As I understand it, there's still an issue where another async task might intervene during a card transaction and cause the card to timeout the command, which we don't currently handle. That would need to be fixed.
> 
> Could this be the reason for the zero length file? I definitely have many other async tasks, pretty complex application.
> 
> If after I write I test the file length and retry in case it's not as expected, will it use the cache for length information or will it check the sd card for real? If cache, any way to clear cache and test?
> 
> > I still also need to decide of the pain of `bisync` is worth the benefit of 'async support' that I don't personally need or use right now. I'd certainly want some embedded async Rust experts on-board for maintenance, because I don't have any way of testing it.
> 
> First, it's working well so far for me and multiple users of my project. Other than two cases of zero length files after write.
> 
> And I don't see how any complex application can be written without async support, maybe there's some technique I'm not aware of but I run tens of tasks simultaneously and without async wouldn't be able to use this library. Just stating that from my POV it's definitely worth any complexity added. Just my 2 cents.

Hi @yanshay - great work on this! Any chance the zero length file was caused by not flushing in time for the directory entry to be updated (data is there but doesn't reflect when you browse the SD Card contents). That had confused me greatly in the past.

zpg6:
> As I understand it, there's still an issue where another async task might intervene during a card transaction and cause the card to timeout the command, which we don't currently handle. That would need to be fixed.

@thejpster can you be more specific about "another async task might intervene" - do you mean two tasks trying to write to the card? Because other activity on SPI should be allowed by other tasks of course "during sequence of a transaction".

thejpster:
* Task A sends command to card
* Task A awaits a small delay
* Task B runs and does a lot of data processing that takes a long time
* Task A resumes and tries to read a response from the card, which has forgotten the original request because it was so long ago

For this library to be generally useful for async and RTOS use-cases, it needs to be robust about retrying commands. I never designed it for those use-cases. I designed it for Arduino style blocking code where any interrupts were always super short.

zpg6:
> * Task A sends command to card
> 
> * Task A awaits a small delay
> 
> * Task B runs and does a lot of data processing that takes a long time
> 
> * Task A resumes and tries to read a response from the card, which has forgotten the original request because it was so long ago
> 
> 
> 
> For this library to be generally useful for async and RTOS use-cases, it needs to be robust about retrying commands. I never designed it for those use-cases. I designed it for Arduino style blocking code where any interrupts were always super short.

Okay that makes more sense. 

When you say "has forgotten the request" - you're referring to the processor in the SDcard itself will have moved on from the transaction after a timeout and has no memory reference of it? Can you point to where in the specification this is described more in detail?

yanshay:
> > > As I understand it, there's still an issue where another async task might intervene during a card transaction and cause the card to timeout the command, which we don't currently handle. That would need to be fixed.
> > 
> > 
> > Could this be the reason for the zero length file? I definitely have many other async tasks, pretty complex application.
> > If after I write I test the file length and retry in case it's not as expected, will it use the cache for length information or will it check the sd card for real? If cache, any way to clear cache and test?
> > > I still also need to decide of the pain of `bisync` is worth the benefit of 'async support' that I don't personally need or use right now. I'd certainly want some embedded async Rust experts on-board for maintenance, because I don't have any way of testing it.
> > 
> > 
> > First, it's working well so far for me and multiple users of my project. Other than two cases of zero length files after write.
> > And I don't see how any complex application can be written without async support, maybe there's some technique I'm not aware of but I run tens of tasks simultaneously and without async wouldn't be able to use this library. Just stating that from my POV it's definitely worth any complexity added. Just my 2 cents.
> 
> Hi @yanshay - great work on this! Any chance the zero length file was caused by not flushing in time for the directory entry to be updated (data is there but doesn't reflect when you browse the SD Card contents). That had confused me greatly in the past.

I always close the directory after writes. My understanding is that this stores the data to the sdcard. Is there a separate flush operation for the directory?

yanshay:
> * Task A sends command to card
> * Task A awaits a small delay
> * Task B runs and does a lot of data processing that takes a long time
> * Task A resumes and tries to read a response from the card, which has forgotten the original request because it was so long ago
> 
> For this library to be generally useful for async and RTOS use-cases, it needs to be robust about retrying commands. I never designed it for those use-cases. I designed it for Arduino style blocking code where any interrupts were always super short.

I can only say from my experience it works pretty well in a pretty complex async application (though load on sdcard is not high but continuous). So maybe it is not far from being robust also for async use cases.

thejpster:
> When you say "has forgotten the request" - you're referring to the processor in the SDcard itself will have moved on from the transaction after a timeout and has no memory reference of it? Can you point to where in the specification this is described more in detail?

I don't think it's mandated in the SD Card specification that a card *must* forget a previous command, but I think it's reasonable for an SD card to be designed that it *will* forget a previous command. After all, it needs to accomodate spurious MCU crashes and reboots and other issues.

robamu:
Hey.

This looks interesting. I'd like to learn lower-level SDMMC and I will probably add embedded-sdmmc support for the Zynq7000 I have here. I think async support would also be very useful for me :) But let's see how far I get with the blocking impl first...

yanshay:
> Hey.
> 
> This looks interesting. I'd like to learn lower-level SDMMC and I will probably add embedded-sdmmc support for the Zynq7000 I have here. I think async support would also be very useful for me :) But let's see how far I get with the blocking impl first...

Just keep in mind that if you are using the main crate branch there have been breaking changes afaik and porting to async may not be straightforward. This branch is behind.
If you use what’s on this branch then you are fine later switching to async.

Be-ing:
I didn't realize there were so many merge conflicts on this branch now. I'm not resolving them.

> I still also need to decide of the pain of bisync is worth the benefit of 'async support' that I don't personally need or use right now.

I'm not going to keep maintaining a branch for months/years while conflicting changes get merged to the develop branch, certainly not without unambiguous support from the maintainer. Been there, done that, not doing it again. This PR demonstrates that using bisync would be a maintainable approach to adding async support to this crate. However, this is a large branch touching the entire crate; it is not easy to maintain out of tree. I'm frankly a bit confused about @thejpster's lack of interest in this, particularly considering how many people have asked for this and contributed towards it. As far as I'm concerned, this was ready to merge months ago, even if there were outstanding issues. It could have been merged with warnings in the documentation, or with a Cargo feature flag that defaulted to off, or not exposing async in the public API; any of those would have been better than letting this branch rot. I'd be glad if someone rebased this branch in a new PR or rewrote it on top of the current develop branch. I wouldn't recommend spending your effort on that without @thejpster agreeing in advance that it would be worth merging, or planning to maintain a fork yourself. Or use [embedded-fatfs](https://github.com/MabezDev/embedded-fatfs/) (which I found out existed after I started this branch).

I started this branch as a proof-of-concept. It was useful for testing the hardware I was working with. It turned out there were other blockers besides async SD card I/O for that project, so I put it on the back burner and am focusing on other projects now. So I have no motivation to deal with merge conflicts with no end in sight.

yanshay:
@Be-ing Sorry to see you go, but thank you very much for the work you've invested in this branch, it has been put to good use.
I had the same experience with at least one other repo I added async to so I keep a fork with the async branch. It seems like when not using async the maintainers have a meaningful barrier to verify the async is working well so they are not fond of merging async stuff. That's a major challenge for the rust embedded community. I actually don't understand how developers develop anything meaningful w/o using async, but I probably miss some basic technique when using blocking to do so.

That said, I think its better, even if you don't plan to invest any more in this, to keep the branch open in case someone needs it. When branch is closed its more difficult to find and the discussion here is very valuable. Anyone who would decide to use it will go over the discussion here and decide if they want to use it or not.

Thanks again!

robamu:
Oh, it definitely was not my intention to see this closed..

> I actually don't understand how developers develop anything meaningful w/o using async

Async Rust seems to be the replacement for a traditional RTOS. Non-blocking C/C++ is either tied to the RTOS (we use a proprietary C FS driver in some other project, which was dependent on FreeRTOS.. blergh) or the API needs to be designed to support blocking (e.g. nb).

> It seems like when not using async the maintainers have a meaningful barrier to verify the async is working well so they are not fond of merging async stuff

It would add to the maintenance burden, and the library was never designed for async (as opposed to embedded-fatfs). I know that smoltcp does not use async, but the low-level layers were explicitely designed to support non-blocking DMA, and there is embassy-net which adds an async API on top of smoltcp. However, it might not be that easy to do this here, I think the low level functions here are methods like read/write which are either blocking or async?

I am not an expert with SDMMC, but I know that the Zynq7000 chip has a dedicated SDIO peripheral and various DMA options.
embedded-fatfs seems nice, but I wonder how well maintained it is seeing that the last commits were merged last year..

yanshay:
> embedded-fatfs seems nice, but I wonder how well maintained it is seeing that the last commits were merged last year..

I just tried it out and managed to get it working with a simple application on esp32s3, both with DMA and without. The API seems easier to use. But I also have concerns it's not really maintained. I saw one of the forks has merged all PR's there a few days ago.  Some of the PR's seems pretty basic, like `alloc` feature not compiling.

Be-ing:
> a dedicated SDIO peripheral

This crate only works with SPI (which, I've found out through hard experience, it not really sufficient for glitch free audio playback). embedded-fatfs seems to be designed to work with different low level protocols, though I think only SPI has been implemented so far?

embedded-fatfs has not been maintained for a while, though the author is a very active member of Espressif's Rust team. Maybe he could be motivated to spend a bit of time on embedded-fatfs maintenance with community encouragement.

thejpster:
> maybe he could be motivated to spend a bit of time on embedded-fatfs maintenance with community encouragement

Please do not attempt to "motivate" maintainers. None of us is getting paid to do this maintenance, and the more PRs and "community engagement" we get, the more free work the maintainers have to do. If you'd like to see changes, send small, clean, well tested PRs that either umabigiously solve known issues, or that add features the maintainer is looking to receive and is happy to maintain in the future.

> SPI ... not really sufficient for glitch free audio playback

I have had no problem streaming 16-bit 48 kHz stereo PCM audio using embedded-sdmmc over SPI. I even did it on stage live during a conference talk. The bandwidth required is 1.536 Mbit/sec, which even on a very slow 5 MHz SPI link should be no problem whatsoever.

Locking this PR because further discussion here is not useful for this crate.

PR #180, opened 2025-02-25
Implement `core::error::Error` for all error types

### LINKED ISSUE #179: Implement `core::error::Error`
Hey, thanks for your work on this nice library!
I think it would be great if the main error type of the crate would implement `core::error::Error`,  thus making it easier to interoperate with other libraries in the ecosystem. However, this would increase the MSRV to 1.81, and also might pull in some of the formatting machinery in `core::fmt`.
If you are open to this I can send in a PR.

michael-p:
Closes #179

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         # Always run MSRV too!
-        rust: ["stable", "1.76"]
+        rust: ["stable", "1.86"]
         features: ['log', 'defmt-log', '""']
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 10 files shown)


michael-p:
> The new feature needs adding to the CI workflow, but it'll break the matrix because you can't build it with 1.76. I don't have any good ideas on how to fix that right now - happy to hear if you have any.

Argh, that's annoying! Sorry, no ideas from my side, except just duplicating the workflow for the `core-error` feature with Rust 1.81. Do you want me to do that, or would you prefer to just not test this in CI?

thejpster:
does https://github.com/orgs/community/discussions/26253#discussioncomment-3250989 help?

I wouldn't want to add something that wasn't being tested in CI, as it'll just break.

avsaase:
You can exclude specific configurations from a matrix job: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude

thejpster:
MSRV was bumped in #204 so a rebase might fix this

thejpster:
Thank you for your PR.

PR #183, opened 2025-02-28
fix bad typing in defmt debug

### LINKED ISSUE #182: Compile error in user code when using defmt-log feature
When writing an application for rp2040 using this library I run into this compile error (which doesn't appear when building the library in its own project).
```
error[E0308]: mismatched types
   --> /home/val/.cargo/git/checkouts/embedded-sdmmc-rs-e7242a4f827ce914/7b15cf5/src/./inner/fat/volume.rs:598:17
    |
598 | /                 debug!(
599 | |                     "LFN Contents {=u8} {=u8} {=u8:02x} {=[?; 13]:#04x}",
600 | |                     start, sequence, csum, buffer
601 | |                 );
    | |                 ^
    | |                 |
    | |_________________expected `&u8`, found `&bool`
    |                   arguments to this function are incorrect
    |
    = note: expected reference `&u8`
               found reference `&bool`
```
I will do a quick PR to solve this, since changing the first `{=u8}` to `{=bool}` solves the problem on a smaller example.

ValouBambou:
This should fix #182.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -27,5 +27,7 @@ jobs:
         toolchain: ${{ matrix.rust }}
     - name: Build
       run: cargo build --no-default-features --features ${{matrix.features}} --verbose
+      env:
+        DEFMT_LOG: debug
     - name: Run Tests
       run: cargo test --no-default-features --features ${{matrix.features}} --verbose
--- a/src/fat/volume.rs
+++ b/src/fat/volume.rs
@@ -594,7 +594,7 @@ impl FatVolume {
                 debug!("LFN Contents {start} {sequence} {csum:02x} {buffer:04x?}");
                 #[cfg(feature = "defmt-log")]
                 debug!(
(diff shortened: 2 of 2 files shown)


thejpster:
Thank you for the fix. Would you be able to amend CI so it actually builds with these logs enabled so this doesn't happen again?

ValouBambou:
I can, but it seems that the `defmt-log` features is already present in the matrix in the [ci yaml file](https://github.com/rust-embedded-community/embedded-sdmmc-rs/blob/5a5b4a9b7f5c8a792276537feb9039651d0b5474/.github/workflows/rust.yml#L22), so I don't know why this error has not being caught before :/

ValouBambou:
Ok I tried to investigate on why this error doesn't appear on the CI, but I cannot reproduce the issue anymore.
That's really weird because when writing this in a lib or a bin (like embassy rp blinky example) it doesn't compile:
```rust
let a = true;
defmt::info!("a = {=u8}", a);
```
Maybe it is a nightly issue (I updated the toolchains several times since I open the issue) but I'm not sure at all.

thejpster (approved):

Issue #184, opened 2025-04-08
Sending `0xFF` byte with CS deasserted when used on shared bus


avsaase:
_This is related to https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/176 but since it also affect the non-async api I figured it deserves it's own issue._

As discussed in https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/176#issuecomment-2781292482, according to several non-official sources, SD cards that share the bus with other slaves need to be sent a dummy `0xFF` byte with CS deasserted after every communication, before communicating with other slaves. Otherwise a bus conflict can occur.

> If you share the SPI bus with other slaves, always send another "dummy" byte (or more) with the CS line inactive (high) after each communication to the SD card before accessing other slaves. Otherwise, the SD card will keep the DO/MISO ((master in slave out)) output active and cause two slaves to drive/short-circuit the MISO line! So: CS low, exchange data with SD card, CS high, send dummy data (e.g. one byte 0xFF), then you can communicate with other slaves.

https://electronics.stackexchange.com/a/602106/256470

> In the SPI bus, each slave device is selected with separated CS signals, and plural devices can be attached to an SPI bus. Generic SPI slave device enables/disables its DO output by CS signal asynchronously to share an SPI bus. However MMC/SDC enables/disables the DO output in synchronising to the SCLK. This means there is a posibility of bus conflict with MMC/SDC and another SPI slave that shares an SPI bus. [...] Therefore to make MMC/SDC release the MISO line, the master device needs to send a byte after the CS signal is deasserted.

http://elm-chan.org/docs/mmc/mmc_e.html#spibus

I brought this up [on matrix](https://matrix.to/#/!BHcierreUuwCMxVqOf:matrix.org/$VLw4fSo5rmLAe9gxC9XKWZ0dg3xro3ckBuq8iT0qLMU?via=matrix.org&via=catircservices.org&via=tchncs.de) and the conclusion was that this requirement is fundamentally not compatible with the `SpiDevice` trait. Workarounds like using a second `SpiDevice` don't solve the problem. This applies to both the blocking and the async versions because in both cases there's no way to guarantee that other tasks or threads don't sneak in and use the bus between the main communication and sending the dummy byte.

[...]

PR #185, opened 2025-04-10
Add SdCardDevice trait

avsaase:
Closes https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/184.

This only adds a blocking trait. Once https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/176 is merged I can rebase and add an sync version. Or this can be merged first and the other PR can add the async version. Either is fine with me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,11 @@ designed for readability and simplicity over performance.
 
 You will need something that implements the `BlockDevice` trait, which can read and write the 512-byte blocks (or sectors) from your card. If you were to implement this over USB Mass Storage, there's no reason this crate couldn't work with a USB Thumb Drive, but we only supply a `BlockDevice` suitable for reading SD and SDHC cards over SPI.
 
+To accommodate specific requirements when using SD cards on a shared SPI bus, this crate uses a bespoke `SdCardDevice` trait instead of the more commmon `embedded_hal::spi::SpiDevice` trait. Implementations for several types that wrap a `embedded-hal::spi::SpiBus` implementation are provided in the `sd_card` module. Some are guarded behind cargo features:
+
+- `embedded-hal-bus-03`: adds support for `embedded-hal-bus::spi::ExclusiveDevice`. This is probably the easiest way to use this crate when the SD card is the only device on the bus.
+- `embassy-sync-06`: adds support for using a blocking mutex from the `embassy-sync` crate. This allows sharing the bus between multiple tasks.
+
 ```rust
 use embedded_sdmmc::{SdCard, VolumeManager, Mode, VolumeIdx};
 // Build an SD Card interface out of an SPI device, a chip-select pin and the delay object
--- a/src/sdcard/sd_card_device.rs
+++ b/src/sdcard/sd_card_device.rs
@@ -1,0 +1,164 @@
+//! SD card device trait and provided implementations.
+
+use core::cell::RefCell;
+
+use embedded_hal::{
+    digital::OutputPin,
+    spi::{Operation, SpiBus},
+};
+
+/// Trait for SD cards connected via SPI.
+pub trait SdCardDevice {
+    /// Perform a transaction against the device.
+    ///
+    /// This is similar to [`embedded_hal::spi::SpiDevice::transaction`], except that this sends
+    /// a dummy `0xFF` byte to the device after deasserting the CS pin but before unlocking the
+    /// bus.
+    fn transaction(
+        &mut self,
+        operations: &mut [Operation<'_, u8>],
+    ) -> Result<(), SdCardDeviceError>;
+
+    /// Do a read within a transaction.
+    ///
+    /// This is a convenience method equivalent to `device.transaction(&mut [Operation::Read(buf)])`.
+    ///
+    /// See also: [`SdCardDevice::transaction`], [`embedded_hal::spi::SpiBus::read`]
+    #[inline]
+    fn read(&mut self, buf: &mut [u8]) -> Result<(), SdCardDeviceError> {
+        self.transaction(&mut [Operation::Read(buf)])
+    }
+
+    /// Do a write within a transaction.
+    ///
+    /// This is a convenience method equivalent to `device.transaction(&mut [Operation::Write(buf)])`.
+    ///
+    /// See also: [`SdCardDevice::transaction`], [`embedded_hal::spi::SpiBus::write`]
+    #[inline]
+    fn write(&mut self, buf: &[u8]) -> Result<(), SdCardDeviceError> {
+        self.transaction(&mut [Operation::Write(buf)])
+    }
+
+    /// Do a transfer within a transaction.
+    ///
+    /// This is a convenience method equivalent to `device.transaction(&mut [Operation::Transfer(read, write)]`.
+    ///
+    /// See also: [`SdCardDevice::transaction`], [`embedded_hal::spi::SpiBus::transfer`]
+    #[inline]
+    fn transfer(&mut self, read: &mut [u8], write: &[u8]) -> Result<(), SdCardDeviceError> {
+        self.transaction(&mut [Operation::Transfer(read, write)])
+    }
+
+    /// Do an in-place transfer within a transaction.
+    ///
+    /// This is a convenience method equivalent to `device.transaction(&mut [Operation::TransferInPlace(buf)]`.
+    ///
+    /// See also: [`SdCardDevice::transaction`], [`embedded_hal::spi::SpiBus::transfer_in_place`]
+    #[inline]
+    fn transfer_in_place(&mut self, buf: &mut [u8]) -> Result<(), SdCardDeviceError> {
+        self.transaction(&mut [Operation::TransferInPlace(buf)])
+    }
+}
+
+/// Errors that can occur when using the [`SdCardDevice`].
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+#[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum SdCardDeviceError {
+    /// An operation on the inner SPI bus failed.
+    Spi,
+    /// Setting the value of the Chip Select (CS) pin failed.
+    Cs,
+}
+
+impl<BUS, CS> SdCardDevice for (&RefCell<BUS>, CS)
+where
+    BUS: SpiBus,
+    CS: OutputPin,
+{
+    fn transaction(
+        &mut self,
+        operations: &mut [Operation<'_, u8>],
+    ) -> Result<(), SdCardDeviceError> {
+        let (bus, cs) = self;
+        let mut bus = bus.borrow_mut();
+        bus_transaction(&mut *bus, cs, operations)
+    }
+}
+
+#[cfg(feature = "embassy-sync-06")]
+impl<CS, BUS, M> SdCardDevice for (&embassy_sync_06::blocking_mutex::Mutex<M, RefCell<BUS>>, CS)
+where
+    CS: OutputPin,
+    BUS: SpiBus,
+    M: embassy_sync_06::blocking_mutex::raw::RawMutex,
+{
+    fn transaction(
+        &mut self,
+        operations: &mut [Operation<'_, u8>],
+    ) -> Result<(), SdCardDeviceError> {
+        let (bus, cs) = self;
+        bus.lock(|bus| {
+            let mut bus = bus.borrow_mut();
+            bus_transaction(&mut *bus, cs, operations)
+        })
+    }
+}
+
+// `ExclusiveDevice` represents exclusive access to the bus so there's no need to send the dummy
+// byte after deasserting the CS pin. We can defer the implementation to the `embedded_hal` trait.
+#[cfg(feature = "embedded-hal-bus-03")]
+impl<CS, BUS, D> SdCardDevice for embedded_hal_bus_03::spi::ExclusiveDevice<BUS, CS, D>
+where
+    BUS: SpiBus,
+    CS: OutputPin,
+    D: embedded_hal::delay::DelayNs,
+{
+    fn transaction(
+        &mut self,
+        operations: &mut [Operation<'_, u8>],
+    ) -> Result<(), SdCardDeviceError> {
+        <Self as embedded_hal::spi::SpiDevice>::transaction(self, operations)
+            .map_err(|_| SdCardDeviceError::Spi)
+    }
+}
+
+fn bus_transaction<BUS, CS>(
+    bus: &mut BUS,
+    cs: &mut CS,
+    operations: &mut [Operation<'_, u8>],
+) -> Result<(), SdCardDeviceError>
+where
+    BUS: SpiBus,
+    CS: OutputPin,
+{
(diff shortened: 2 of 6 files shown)


src/sdcard/sd_card_device.rs
```diff
@@ -0,0 +1,164 @@
...
+#[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
+#[non_exhaustive]
+pub enum SdCardDeviceError {
+    /// An operation on the inner SPI bus failed.
+    Spi,
+    /// Setting the value of the Chip Select (CS) pin failed.
+    Cs,
+}
+
+impl<BUS, CS> SdCardDevice for (&RefCell<BUS>, CS)
```
avsaase, line 74: I couldn't add an impl for `(BUS, CS) because it conflicted with this one.

src/sdcard/sd_card_device.rs
```diff
@@ -0,0 +1,213 @@
...
+    let dummy_res = bus.write(&[0xFF]);
+    let flush_res = bus.flush();
+
+    dummy_res.map_err(|_| SdCardDeviceError::Spi)?;
+    flush_res.map_err(|_| SdCardDeviceError::Spi)?;
+
+    Ok(())
+}
+
+fn send_clock_pulses<BUS, CS>(bus: &mut BUS, cs: &mut CS) -> Result<(), SdCardDeviceError>
```
thejpster, line 198: Missing docs

src/sdcard/sd_card_device.rs
```diff
@@ -0,0 +1,213 @@
...
+        operations: &mut [Operation<'_, u8>],
+    ) -> Result<(), SdCardDeviceError> {
+        <Self as embedded_hal::spi::SpiDevice>::transaction(self, operations)
+            .map_err(|_| SdCardDeviceError::Spi)
+    }
+
+    fn send_clock_pulses(&mut self) -> Result<(), SdCardDeviceError> {
+        let bus = self.bus_mut();
+
+        // There's no way to access the CS pin here so we can't set it high. Most like it already high so this is probbaly fine(?)
```
thejpster, line 145:
```suggestion
        // There's no way to access the CS pin here so we can't set it high. Most likely it is already high so this is probably fine(?)
```

thejpster:
Seems reasonable enough, but I'd like to see some currently users of embedded-sdmmc-rs port over to this style to check it's usable in practice.

avsaase:
> Seems reasonable enough, but I'd like to see some current users of embedded-sdmmc-rs port over to this style to check it's usable in practice.

I'll also do some dogfooding by using it in my project. I'm mostly interested in the async API but I think I can do some tests with the block API.

> There's also some overlap with #170, which is about SD Cards in general (i.e. using SD Card Host Controllers with the 4-bit bus) rather than this which is about SPI bus based SD Cards specifically.

How do you want to move forward with this? There hasn't been any recent activity on that PR. I think it makes sense to first fix the dummy byte issue.

thejpster:
If we're going to abstract the SD card transport it makes sense to do it in a way that both fixes the SPI issue and allows the use of SD Host Controllers.

I also have a mild preference for custom types over implementing traits on tuples. I think it makes it clearer what's going on.

Maybe something like:

* struct VolumeManager uses a BlockDevice
* SdCard implements BlockDevice using an SdCardTransport
* RefCellSpiSdCardTransport implements SdCardTransport using a &RefCell<SpiBus>
* Stm32SdHostTransport also implements SdCardTransport
* etc

avsaase:
Personally I don't have the need for a generic `Transport` trait. I understand why it would be useful to have but I think that's a separate thing from what this PR tries to a do. If @luojia65 is still interested in their PR I'm happy to discuss how we can combine efforts but I would like to limit the scope of this PR to just fixing the dummy byte issue.

README.md
```diff
@@ -10,6 +10,11 @@ designed for readability and simplicity over performance.
...
 
+To accommodate specific requirements when using SD cards on a shared SPI bus, this crate uses a bespoke `SdCardDevice` trait instead of the more commmon `embedded_hal::spi::SpiDevice` trait. Implementations for several types that wrap a `embedded-hal::spi::SpiBus` implementation are provided in the `sd_card` module. Some are guarded behind cargo features:
+
+- `embedded-hal-bus-03`: adds support for `embedded-hal-bus::spi::ExclusiveDevice`. This is probably the easiest way to use this crate when the SD card is the only device on the bus.
+- `embassy-sync-06`: adds support for using a blocking mutex from the `embassy-sync` crate. This allows sharing the bus between multiple tasks.
+
 ```rust
 use embedded_sdmmc::{SdCard, VolumeManager, Mode, VolumeIdx};
 // Build an SD Card interface out of an SPI device, a chip-select pin and the delay object
```
thejpster, line 20: This example needs to be kept in sync with the copy in readme-test.

thejpster:
Ok, fair enough.

ValouBambou:
Is this still planned? I wrote a patch to fix the build test checks and open a PR in @avsaase fork.

avsaase:
Some (positive) life events happened so I couldn't work on this for a few weeks but I'm still planning on getting this merged. The last thing I did was to quickly whip up an async version of this PR to see how that would work. Unfortunately it did not work at all and consistently gave CRC errors. I haven't had the chance yet to see if this was a problem with this PR or my quick async version of it (it doesn't look like I pushed that code to github). In any case it's probably best to test the code from this PR first to see if the basic approach is correct. If you want to help with that that would be great.

ValouBambou:
I could test your 2 versions async and blocking next week if you want I would have access to real hardware (rp2040 + some SD cards). I will just need the async code. I'm not sure what is the best way to include the async code since the other PR is still waiting.

avsaase:
I pushed my async compatible version of this PR here: https://github.com/avsaase/embedded-sdmmc-rs/tree/bisync-sd-card-device.

ValouBambou:
When writing a simple example of just raw reads and writes, I struggle a bit to see how to set the frequency to a high value after the initialization of 400 kHz. We could do this in the previous API : 
```rust
let mut config = embassy_rp::spi::Config::default();
config.frequency = 16_000_000;
sdcard.spi(|dev| dev.bus_mut().set_config(&config));
```
But with the new one I don't see a way of doing that, maybe we should add a bus_mut method to the `SdCardSpiDevice` trait in order to have a mutable borrow to the underlying bus since it's private.

ValouBambou:
I therefore tried with an `ExclusiveDevice` but I realized that it doesn't implement the trait in async. In order to implement it, we need to add the `async` feature from [embedded-hal-bus](https://docs.rs/embedded-hal-bus/latest/src/embedded_hal_bus/spi/exclusive.rs.html#108-153) (since `ExclusiveDevice` implement both async and blocking traits) and some await point should be enough.

avsaase:
> I struggle a bit to see how to set the frequency to a high value after the initialization of 400 kHz

Yeah that's a tricky one. Providing mutable access to the bus does not allow using different SPI speeds for different SPI devices. In my own project that uses embassy I created a newtype and implemented `SdCardSpiDevice` for it. This implementation can then set the SPI speed to the correct value for the SD card device.

```rust
pub struct SdCardSpiDeviceWithConfig<'a, BUS: SetConfig, CS, M: RawMutex> {
    bus: &'a Mutex<M, BUS>,
    cs: CS,
    config: BUS::Config,
}

impl<'a, BUS: SetConfig, CS: OutputPin, M: RawMutex> SdCardSpiDeviceWithConfig<'a, BUS, CS, M> {
    pub fn new(bus: &'a Mutex<M, BUS>, cs: CS, config: BUS::Config) -> Self {
        Self { bus, cs, config }
    }
}

impl<BUS: SpiBus + SetConfig, CS: OutputPin, M: RawMutex> SdCardSpiDevice
    for SdCardSpiDeviceWithConfig<'_, BUS, CS, M>
{
    async fn transaction(&mut self, operations: &mut [Operation<'_, u8>]) -> Result<(), SdCardDeviceError> {
        let mut bus = self.bus.lock().await;
        bus.set_config(&self.config).map_err(|_| SdCardDeviceError::Spi)?;
        bus_transaction(&mut *bus, &mut self.cs, operations).await
    }

    async fn send_clock_pulses(&mut self) -> Result<(), SdCardDeviceError> {
        let mut bus = self.bus.lock().await;
        send_clock_pulses(&mut *bus, &mut self.cs).await
    }
}
```

I just pushed a commit to this branch that makes `bus_transaction` and `send_clock_pulses` public to make this possible. The `embedded-hal` trait does not expose functions to set SPI speeds so I think this is the only way to do it.

marcfir:
> I pushed my async compatible version of this PR here: https://github.com/avsaase/embedded-sdmmc-rs/tree/bisync-sd-card-device.

I tested this version on a ESP32S3 with embassy
```Rust
    let sclk = peripherals.GPIO12;
    let miso = peripherals.GPIO11;
    let mosi = peripherals.GPIO13;
    let cs = peripherals.GPIO10;

    let mut sd_cs = Output::new(cs, Level::High);
    let dma_channel = peripherals.DMA_CH0;

    let (rx_buffer, rx_descriptors, tx_buffer, tx_descriptors) = dma_buffers!(32000);
    let dma_rx_buf = DmaRxBuf::new(rx_descriptors, rx_buffer).unwrap();
    let dma_tx_buf = DmaTxBuf::new(tx_descriptors, tx_buffer).unwrap();

    let mut spi_bus = Spi::new(
        peripherals.SPI2,
        spi::master::Config::default()
            .with_frequency(400.kHz())
            .with_mode(spi::Mode::_0),
    )
    .unwrap()
    .with_sck(sclk)
    .with_mosi(mosi)
    .with_miso(miso)
    .with_dma(dma_channel)
    .with_buffers(dma_rx_buf, dma_tx_buf)
    .into_async();
    let spi_dev = RefCellSdCardDevice::new(spi_bus, sd_cs);
    let sdcard = SdCard::new(spi_dev, Delay);
    sdcard.init_card().await;
    let mut volume_mgr = VolumeManager::new(sdcard, DummyTimesource::default());
    static VOL_MGR: StaticCell<
        VolumeManager<
            SdCard<RefCellSdCardDevice<'_, SpiDmaBus<'_, esp_hal::Async>, Output<'_>>, Delay>,
            DummyTimesource,
        >,
    > = StaticCell::new();
    let volume_mgr = VOL_MGR.init(volume_mgr);
    let volume0 = volume_mgr.open_volume(VolumeIdx(0)).await.unwrap();
```

It crashes on this `    let volume0 = volume_mgr.open_volume(VolumeIdx(0)).await.unwrap();` line with
```
DEBUG - Creating new embedded-sdmmc::VolumeManager
TRACE - Reading partition table
DEBUG - Read 1 blocks @ 0
DEBUG - acquiring card with opts: AcquireOpts { use_crc: true, acquire_retries: 50 }
TRACE - Reset card..
TRACE - Enter SPI mode, attempt: 1..
WARN - Got response: 3f, trying again..
TRACE - Enter SPI mode, attempt: 2..
DEBUG - Enable CRC: true
DEBUG - Card version: SDHC
ERROR - read: ReadError


====================== PANIC ======================
panicked at soundcube\src\main.rs:229:62:
called `Result::unwrap()` on an `Err` value: DeviceError(ReadError)
```
When analyzing the bus I see CMD17 as the last command. Is the CS line correct?
image

I think https://github.com/avsaase/embedded-sdmmc-rs/blob/08fc93c092da0c030e9618264c805d0fedea669b/src/inner/sdcard/mod.rs#L371-L373 is where the error is thrown?

ValouBambou:
Maybe I get something wrong, but I can't see why just exposing a method to get a mutable reference to the underlying `SpiBus` in `RefCellSdCardDevice` or in the trait `SpiCardSpiDevice`directly is not an option. It's less boilerplate than your custom type and just depend on if the underlying type supports to change the frequency or not. I tried to add such method in `RefCellSdCardDevice` and it worked well on my clone. 

```rust
impl<'a, BUS, CS> RefCellSdCardDevice<'a, BUS, CS> {
    /// Try stuff
    pub fn foo(&self) -> RefMut<'_, BUS> {
        self.bus.borrow_mut()
    }
}
 ```

However, I also confirmed running read errors on a different target (rp2040).
I didn't have this issue with an older commit (I will try to bisect by hand since API are changed with different commits). Here is the main code for reproduction :
```rust
#![no_std]
#![no_main]
#![feature(impl_trait_in_assoc_type)]

use core::cell::RefCell;

use defmt::*;
use embassy_executor::Spawner;
use embassy_rp::gpio;
use embassy_rp::spi::Spi;
use embedded_sdmmc::asynchronous::{SdCard, sdcard::device::RefCellSdCardDevice};
use gpio::{Level, Output};
use {defmt_rtt as _, panic_probe as _};

#[embassy_executor::main]
async fn main(_spawner: Spawner) {
    // BEGIN stuff with real peripherals
    let p = embassy_rp::init(Default::default());
    let mut config = embassy_rp::spi::Config::default();
    config.frequency = 400_000;
    let spi_bus = Spi::new(
        p.SPI1, p.PIN_10, p.PIN_11, p.PIN_12, p.DMA_CH0, p.DMA_CH1, config,
    );
    let cs = Output::new(p.PIN_13, Level::High);
    // END stuff with real peripherals
    let delay = embassy_time::Delay;
    let bus = RefCell::new(spi_bus);
    let spi_device = RefCellSdCardDevice::new(&bus, cs);
    let sdcard = SdCard::new(spi_device, delay);
    sdcard.init_card().await.unwrap();
    info!("init is ok");
    let mut config = embassy_rp::spi::Config::default();
    config.frequency = 16_000_000;
    sdcard.spi(|dev| dev.foo().set_config(&config));
    info!("freq is ok");
    info!("Card size is {} bytes", sdcard.num_bytes().await.unwrap());
}
```

avsaase:
> Maybe I get something wrong, but I can't see why just exposing a method to get a mutable reference to the underlying `SpiBus` in `RefCellSdCardDevice` or in the trait `SpiCardSpiDevice`directly is not an option. It's less boilerplate than your custom type and just depend on if the underlying type supports to change the frequency or not. I tried to add such method in `RefCellSdCardDevice` and it worked well on my clone.

How would you use two devices with differett speeds on the same bus? You need a way to change the speed for every bus transaction and I think the only way to do that properly is in the `SdCardSpiDevice` trait impl. This is similar to what embassy does with their `SpiDeviceWithConfig` type.

> EDIT: I found the bad commit is this one [835b2e4](https://github.com/rust-embedded-community/embedded-sdmmc-rs/commit/835b2e4f9d3482b6287f674d7ecf6ae5d0618c18) (i.e the last one pushed on the [async fork](https://github.com/Be-ing/embedded-sdmmc-rs/tree/bisync))

That's strange. I and several of my users are using that exact commit without any problem for weeks now.

ValouBambou:
> How would you use two devices with differett speeds on the same bus? You need a way to change the speed for every bus transaction and I think the only way to do that properly is in the `SdCardSpiDevice` trait impl. This is similar to what embassy does with their `SpiDeviceWithConfig` type.

Yes, and you can call the set_config method directly, like in my example, by just adding a tiny get_inner method. I don't see why you need to reimplement the trait "only for that". If you need to set the frequency before each transaction you will reimplement it for your custom type to avoid calling get_inner every time, but for using it once it's easier with just exposing the reference to the concrete type.

Issue #188, opened 2025-05-30
Bug? Can't Seek to offset 0 from start and write - file gets corrupted


yanshay:
I'm writing to a file in chunks of 128 bytes. Every write is open file, write, flush and close of the file. Each such write I'll refer to as a transaction.

After writing three such transactions, the file size is 384 bytes.

Then
- If I seek to position 128 and write 128 bytes - write is ok, and those 128 bytes are modified.
- If I seek to position 256 and write 128 bytes - write is ok, and those 128 bytes are modified.
- If I seek to position 0 and write 128 bytes - write succeeds, but file is corrupted. First 128 bytes are as expected, the following 256 are zeros.
- If I don't seek it writes at the end of the file, so currently I'm unable to write to the beginning of the file.

I'm using the async (using bisync api), but I don't think it has anything to do with that.

Is it a bug or maybe something I could be doing wrong?

Some technical details in case it's relevent.
I create the file using `ReadWriteCreateOrAppend` and then open using `ReadWriteAppend`.
I use `seek_from_start`, `write`, `flush`, `close` - everything pretty standard.

thejpster:
Away at the moment, but sounds plausible. I'd love to see a PR with a test case that trips the bug (and hence fails) followed by a commit that demonstrably fixes it. The PR should be on the main branch - we can rebase the async stuff after.

thejpster:
I commented in the PR, but for anyone who finds this issue, my proposed fix is:

```rust
let block = if (block_offset == 0) && (to_copy == Block::LEN) {
    // we're replacing the whole Block, so the previous contents
    // are irrelevant
    data.block_cache.blank_mut(block_idx)
} else {
    debug!("Reading for partial block write");
    data.block_cache
        .read_mut(block_idx)
        .map_err(Error::DeviceError)?
};
```

i.e. always read the current contents of a block unless we know we are replacing the whole block anyway.

thejpster:
Closed by #189

PR #189, opened 2025-06-07
fix seek and write to block start zeros rest of block after written data (#188)

yanshay:
Fix for #188 where seeking to offset of block size multiple and then writing, writes the data and zero the leftover of the block size instead of keeping the data that was there.
Includes also test for seek+write at both offset that worked and offset that didn't work.

### DIFF
--- a/src/volume_mgr.rs
+++ b/src/volume_mgr.rs
@@ -878,7 +878,9 @@ where
                 Err(e) => return Err(e),
             };
             let to_copy = core::cmp::min(block_avail, bytes_to_write - written);
-            let block = if block_offset != 0 || data.open_files[file_idx].current_offset < data.open_files[file_idx].entry.size {
+            let block = if block_offset != 0
+                || data.open_files[file_idx].current_offset < data.open_files[file_idx].entry.size
+            {
                 debug!("Reading for partial block write");
                 data.block_cache
                     .read_mut(block_idx)
--- a/src/volume_mgr.rs
+++ b/src/volume_mgr.rs
@@ -878,15 +878,15 @@ where
                 Err(e) => return Err(e),
             };
             let to_copy = core::cmp::min(block_avail, bytes_to_write - written);
-            let block = if block_offset != 0
-                || data.open_files[file_idx].current_offset < data.open_files[file_idx].entry.size
-            {
+            let block = if (block_offset == 0) && (to_copy == block_avail) {
+                // we're replacing the whole Block, so the previous contents
+                // are irrelevant
+                data.block_cache.blank_mut(block_idx)
+            } else {
                 debug!("Reading for partial block write");
                 data.block_cache
                     .read_mut(block_idx)
                     .map_err(Error::DeviceError)?
-            } else {
-                data.block_cache.blank_mut(block_idx)
             };
             block[block_offset..block_offset + to_copy]
                 .copy_from_slice(&buffer[written..written + to_copy]);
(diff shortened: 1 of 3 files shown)


src/volume_mgr.rs
```diff
@@ -878,7 +878,9 @@ where
                 Err(e) => return Err(e),
             };
             let to_copy = core::cmp::min(block_avail, bytes_to_write - written);
-            let block = if block_offset != 0 || data.open_files[file_idx].current_offset < data.open_files[file_idx].entry.size {
+            let block = if block_offset != 0
```
thejpster, line 881: I think I would prefer something like:

```rust
            let block = if (block_offset == 0) && (to_copy == Block::LEN) {
                // we're replacing the whole Block, so the previous contents
                // are irrelevant
                data.block_cache.blank_mut(block_idx)
            } else {
                debug!("Reading for partial block write");
                data.block_cache
                    .read_mut(block_idx)
                    .map_err(Error::DeviceError)?
            };
```

This also passes your new test case.

thejpster: I believe that your fix would mean a write of 512 bytes at the start of the file would do a read-modify-write, when a write is sufficient.

yanshay: Will this work also at EOF write happens to be exactly on block boundary and less than Block::LEN?
In that case won't there be in the block_cache potentially data from previous reads? I thought that was the reason for filling blank in the first place, to support appending and clearing data if EOF was right at the beginning of a new block.

thejpster: So actually, I could be persuaded to make this `if (block_offset == 0) && (to_copy == block_avail)`.

thejpster: So I replied to your erased comment. Deleting my replies.

yanshay: I don't have any preference, not really very familiar with the code. Let me know your preference and I'll modify it, I trust your judgement better than mine here.

thejpster: `to_copy` is the amount of stuff we are writing to this block (see the slice copy that happens after). `block_avail` is how much stuff in this block (starting at the beginning of the block, because you cannot have holes in files when using FAT) is actually valid.

This might be better:

```rust
            let block = if (block_offset == 0) && (to_copy == block_avail) {
                // we're replacing the whole Block, so the previous contents
                // are irrelevant
                data.block_cache.blank_mut(block_idx)
            } else {
                debug!("Reading for partial block write");
                data.block_cache
                    .read_mut(block_idx)
                    .map_err(Error::DeviceError)?
            };
```

thejpster: Consider a file that is 1026 bytes long. We write 514 bytes starting at offset 511.

We should:

1. Read-modify-write Block 0 in the file, replacing `block[511]`
2. Write Block 1 in the file.
3. Read-modify write Block 2 in the file, replacing `block[0]` but leaving `block[1]` unchanged (and leaving `block[2..512]` unchanged as well).

Consider a file that is 1026 bytes long. We write 515 bytes starting at offset 511.

We should:

1. Read-modify-write Block 0 in the file, replacing `block[511]`
2. Write Block 1 in the file
3. Write Block 2 - it contains two data bytes at `block[0..2]` and 510 zero bytes after that, but the zeroes are beyond the EOF.

yanshay: I made the update based on the above and also updated the changelog. 
Anything else required?

thejpster (approved):

PR #196, opened 2025-06-30
Fix missing type parameter in VolumeManager::device

### LINKED ISSUE #195: VolumeManager::device expects time source as return value
The [new API](https://github.com/rust-embedded-community/embedded-sdmmc-rs/blob/b97b4f1b1c408849d32b6587ff10a81f3360734c/src/volume_mgr.rs#L97) for temporarily accessing the block device expects the given closure to return something of type `T`:

```rust
/// Temporarily get access to the underlying block device.
pub fn device<F>(&self, f: F) -> T
where
    F: FnOnce(&mut D) -> T,
{
    let mut data = self.data.borrow_mut();
    let result = f(data.block_cache.block_device());
    result
}
```

I am pretty sure that `T` in this case is supposed to be an arbitrary type. However, the `VolumeManager::device` method does not declare `T`, it reuses the type parameter from the struct definition, which is the time source. This makes it hard to actually use this method, since it is expecting to return an instance of the time source.

fko-kuptec:
Fixes #195

### DIFF
--- a/src/volume_mgr.rs
+++ b/src/volume_mgr.rs
@@ -94,9 +94,9 @@ where
     }
 
     /// Temporarily get access to the underlying block device.
-    pub fn device<F>(&self, f: F) -> T
+    pub fn device<R, F>(&self, f: F) -> R
     where
-        F: FnOnce(&mut D) -> T,
+        F: FnOnce(&mut D) -> R,
     {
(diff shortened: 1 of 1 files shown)


thejpster (approved):

thejpster:
Thank you!

PR #198, opened 2025-08-05
Remove the create-dir feature from the todo list

kasper64:
As far as I can see, this feature seems to be there. It's in the docs and there are tests for it. Was the removal from the list forgotten or is it not stable enough?

Kind regards and many thanks for this library,
Kasper


thejpster (approved):

thejpster:
Just an oversight. Thanks!

PR #201, opened 2025-08-28
Fix stack overflow when receiver is typed as embedded_io trait

tmpfs:
Prevents a stack overflow when the receiver is typed as embedded_io::Read or embedded_io::Write.


thejpster (approved):

thejpster:
Thanks!

PR #204, opened 2025-09-15
Update to `defmt` 1.0.1, `embedded-hal-bus` 0.3.0, `env_logger` 0.11.8, `heapless` 0.9.1, and `hex-literal` 1.0.0.

ColinFinck:
The upgrade from `heapless` 0.8.x to 0.9.x is particularly important for me, because I'm heavily using `heapless` in an embedded software project and need to use a single consistent version for interoperability. I took this opportunity to upgrade all other outdated dependencies too.

This raises the MSRV to 1.86.0. Testing has shown that this is the minimum required version for `heapless` 0.9.1.

CC @thejpster

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         # Always run MSRV too!
-        rust: ["stable", "1.76"]
+        rust: ["stable", "1.86"]
         features: ['log', 'defmt-log', '""']
     steps:
     - uses: actions/checkout@v4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog] and this project adheres to [Semantic 
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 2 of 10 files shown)


thejpster:
Ah, GitHub - you suck. Why can’t it work out that the jobs I told it should pass no longer exist in the repo? Why isn’t the pass/fail config in the repo where I could change it rather than living in the database where there is now a race hazard?

I will merge this when I get back to my laptop (I’m at Oxidize) and fix the required jobs setting.

thejpster (approved):

thejpster:
Thank you for your PR.

rursprung:
> Ah, GitHub - you suck. Why can’t it work out that the jobs I told it should pass no longer exist in the repo? Why isn’t the pass/fail config in the repo where I could change it rather than living in the database where there is now a race hazard?

that's why i started using a single `build-results` job in all my workflows: https://github.com/rust-embedded-community/tb6612fng-rs/blob/master/.github/workflows/CI.yml#L98-L112
then you just depend on that and can change everything in the file.

thejpster:
Yes, cortex-ar and the ferrous trainings do something similar. I just haven't updated this one yet.

PR #205, opened 2025-09-20
docs: adds esp32 example

zpg6:
Adding an example of using on ESP32 with esp-hal (no_std) + Embassy: https://github.com/zpg6/esp32-sdcard

- Uses `esp-hal@1.0.0`
- Includes instructions for installing the `esp-rs` toolchain
- Gives some example Micro SD Card Reader/Adapter modules you can buy
- Shows how to format Micro SD Cards with Rufus
- Gives example debug and CSV file outputs created by the example


thejpster (approved):

thejpster:
Thank you!

PR #208, opened 2025-10-31
Implement Ord and PartialOrd for ShortFileName.

de-vri-es:
Can be useful to directly compare short filenames lexicographically.

Should be fine to derive because all characters lower than a space are illegal, and spaces indicate the end of a string. No legal character compares lower than a space, so shorter strings are always sorted before longer string.


thejpster (approved):

thejpster:
Looks good to me. I appreciate the test cases.

de-vri-es:
Thanks for the quick review!

PR #209, opened 2025-11-08
Async support

BL-CZY:


thejpster:
Thank you for this PR. 

Unfortunately this is a big PR with no tests and no explanation of what it does or why it does it.

We’re going to need a bunch of examples and documentation updates before this can go in. If you’re asking me to take on ongoing maintenance for an API I have no need for, you’ll need to convince me it’s correct first and that I’m not just opening myself up for a slew of bug reports.

I’m also going to insist that the async feature is off by default.

BL-CZY:
Hello!

I am sorry for this PR. I will keep working on the branch, adding the tests, and fixing clippy warnings.

thejpster:
I don't mind seeing the PR left open as a draft! Feel free to @ me when you think you've got good test coverage ... which I guess you'd do with Tokio? Not sure.

PR #210, opened 2025-12-18
Support directory deletion

### LINKED ISSUE #186: Implement rmdir
Currently it's possible to remove only files.

https://docs.rs/embedded-sdmmc/0.8.1/embedded_sdmmc/struct.Directory.html#method.delete_file_in_dir

We are missing a way to allow of directory deletion.

estebank:
Fix https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/186.

The changes to the `Error` enum would be a breaking change. The new variant can either be only added, leaving behind a variant that will never be constructed or replaced.

### DIFF
--- a/src/volume_mgr.rs
+++ b/src/volume_mgr.rs
@@ -666,26 +666,62 @@ where
         let data = data.deref_mut();
 
         let dir_idx = data.get_dir_by_id(directory)?;
-        let dir_info = &data.open_dirs[dir_idx];
-        let volume_idx = data.get_volume_by_id(dir_info.raw_volume)?;
+        let parent_dir_info = &data.open_dirs[dir_idx];
+        let volume_idx = data.get_volume_by_id(parent_dir_info.raw_volume)?;
         let sfn = name.to_short_filename().map_err(Error::FilenameError)?;
 
         let dir_entry = match &data.open_volumes[volume_idx].volume_type {
-            VolumeType::Fat(fat) => fat.find_directory_entry(&mut data.block_cache, dir_info, &sfn),
+            VolumeType::Fat(fat) => {
+                fat.find_directory_entry(&mut data.block_cache, parent_dir_info, &sfn)
+            }
         }?;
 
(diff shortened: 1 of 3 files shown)


src/volume_mgr.rs
```diff
@@ -665,26 +665,64 @@ where
...
-        }
-
-        if data.file_is_open(dir_info.raw_volume, &dir_entry) {
+            // Find the directory to be deleted, so that we can check its contents.
+            let dir_info = if let Some(dir_info) = data
+                .open_dirs
+                .iter()
+                .find(|dir_info| {
+                    // Subdirectory is already open.
+                    dir_info.cluster == dir_entry.cluster
```
thejpster: I think that if the directory you are deleting is already open, you shouldn't be able to delete it because it would leave a dangling entry.

src/volume_mgr.rs
```diff
@@ -665,26 +665,64 @@ where
         let data = data.deref_mut();
 
         let dir_idx = data.get_dir_by_id(directory)?;
-        let dir_info = &data.open_dirs[dir_idx];
-        let volume_idx = data.get_volume_by_id(dir_info.raw_volume)?;
+        let parent_dir_info = data.open_dirs[dir_idx].clone();
```
thejpster: I don't recall needing to clone an entry for an open directory before (unless I'm forgetting about it). What's the reason for the clone as opposed to borrowing the entry, or referring to it by index? This is probably fine, it just gave me pause as I read it.

estebank: This is only *needed* if we want to allow deletion of currently open directories, because I'm creating a new `DirectoryInfo` in a block and making it a borrow would require using some of the lifetime extension tricks, while the struct looked tiny enough that cloning seemed cleaner to do.

thejpster:
Rebasing after #211 is merged should fix CI.

estebank:
Done and added test.

thejpster (approved):

thejpster:
Thanks!

PR #211, opened 2025-12-18
Fix CI - bump MSRV to 1.87

thejpster:
* Bumps MSRV to 1.87 to avoid failure seen in https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/210
* Removes core-error feature, because MSRV is high enough we can just always turn it on

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         # Always run MSRV too!
-        rust: ["stable", "1.86"]
+        rust: ["stable", "1.87"]
         features: ['log', 'defmt-log', '""']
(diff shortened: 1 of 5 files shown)


jannic:
Wouldn't it be more robust to set `resolver.incompatible-rust-versions="fallback"` (or `resolver = "3"` or `edition = '2024'
` which changed the default to `fallback`)?

That way, cargo will automatically select versions of dependencies that are compatible with the current compiler.
Bumping the rustc version used in CI only helps with the new MSRV of heapless, but will break again once any dependency updates to a more recent MSRV.

jannic (approved):
While setting `resolver="3"` may still be a good idea, there's nothing wrong with this PR. Approved.

PR #213, opened 2025-12-19
Update to edition 2024

### LINKED ISSUE #212: Wouldn't it be more robust to set `resolver.incompatible-rust-versions="fallback"`
Wouldn't it be more robust to set `resolver.incompatible-rust-versions="fallback"` (or `resolver = "3"` or `edition = '2024'
` which changed the default to `fallback`)?

That way, cargo will automatically select versions of dependencies that are compatible with the current compiler.
Bumping the rustc version used in CI only helps with the new MSRV of heapless, but will break again once any dependency updates to a more recent MSRV.

_Originally posted by @jannic in https://github.com/rust-embedded-community/embedded-sdmmc-rs/issues/211#issuecomment-3676272704_

jannic:
To adapt to new formatting defaults and avoid clippy warnings, this includes changes made by `cargo fmt` and `cargo clippy --fix`.

Overall, I think the clippy fixes are an improvement. I'm not 100% sure about `clippy::manual_range_contains`. Which one is more readable?

```
    (true, sequence, _) if sequence >= 0x02 && sequence < 0x14 => {
```
or
```
    (true, sequence, _) if (0x02..0x14).contains(&sequence) => {
```

To me the old version looks more familiar. But I tend to agree with the reasoning given at https://rust-lang.github.io/rust-clippy/beta/index.html#manual_range_contains. In case there is no consensus on that change, we could also disable that lint instead.

This also fixes #212.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog] and this project adheres to [Semantic 
 - Updated to `defmt` 1.0.1, `embedded-hal-bus` 0.3.0, `env_logger` 0.11.8, `heapless` 0.9.1, and `hex-literal` 1.0.0.
 - Raised the minimum supported Rust version to 1.87.0.
 - Removed `core-error` feature as MSRV is now above 1.81
+- Set edition to 2024
 
 ## [Version 0.9.0] - 2025-06-08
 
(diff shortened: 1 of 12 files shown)


thejpster (approved):
Looks good to me. Nice improvements, and I’m fine with the range::contains syntax too.

PR #216, opened 2025-12-22
Documentation updates

thejpster:
Updates documentation for `VolumeManager`, `Directory` and `File` to add missing details.

Also removes an unused `Error` variant, and renames `delete_file_in_dir` to `delete_entry_in_dir`.

This is a breaking change.

PR #217, opened 2025-12-22
Support opening files by long file name

thejpster:
Adds a new API which lets you open files by their long file name.

Does not (currently) let you open directories in this manner, nor does it let you create files with a long file name.

Builds on #216


thejpster:
rebased

eldruin (commented):
Sounds good to me, thanks!
Could you add an entry to the changelog?
(I rebased this with the GH UI already)

thejpster:
Done

eldruin (approved):
Thank you so much!

wisp3rwind:
Regarding the change to the directory iterator, maybe the callback should be expected to return the standard [`ControlFlow`](https://doc.rust-lang.org/stable/core/ops/enum.ControlFlow.html) type rather than a custom type?

thejpster:
Oh! That’s new. Yes, that’s a better choice.

PR #218, opened 2026-01-22
Reduce `LfnBuffer` size from 24 to 16 bytes on 32-bits sys.

Conaclos:
- Reduce `LfnBuffer` size from 24 to 16 bytes on 32-bits sys.
- Reduce `LfnBuffer` size from 32 to 24 bytes on 64-bits sys.

The reduction comes from the observation that the required buffer doesn't need to exceed the maximum bytes required to represent a LFN. A LFN has at most 255 UTF-16 characters. The buffer holds UTF-8 characters. A UTF-8 character takes at most 3 bytes. So, the buffer doesn't need to exceed `255*3` bytes. Thus, `free` can be stored in a u16.

To avoid a breaking change, the buffer is sliced in order to retain at most `u16::MAX` bytes.

Note that the optimization could go beyond, however this could become too complex IMO.

### DIFF
--- a/src/filesystem/filename.rs
+++ b/src/filesystem/filename.rs
@@ -228,7 +228,7 @@ pub struct LfnBuffer<'a> {
     /// How many bytes are free.
     ///
     /// This is also the byte index the string starts from.
-    free: usize,
+    free: u16,
     /// Did we overflow?
     overflow: bool,
     /// If a surrogate-pair is split over two directory entries, remember half of it here.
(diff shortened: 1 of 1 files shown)


thejpster:
Thank you for the PR. Looking at this, I think I can just about justify the complexity for the memory saving. But perhaps you could add a function like:

```
fn free(&self) -> usize {
    usize::from(self.free)
}
```

That would avoid the need to do so many u16 to usize conversions in the code.

Have you found the eight byte saving to be significant in your application?

Conaclos:
> Have you found the eight byte saving to be significant in your application?

The gain is not significative at an application level. However, I thought it was a good tradeoff between complexity/gain. If you think it is not worth the complexity, feel free to reject it :)

thejpster (approved):

PR #219, opened 2026-01-29
Switch to ControlFlow, plus other small fixes

thejpster:
* Switch to using `core::ops::ControlFlow` instead of my `enum Continue`
* Add some timestamp tests
* Add some missing docs
* Apply some clippy suggestions

### DIFF
--- a/examples/list_dir.rs
+++ b/examples/list_dir.rs
@@ -34,6 +34,8 @@
 mod linux;
 use linux::*;
 
+use std::ops::ControlFlow;
+
 use embedded_sdmmc::{ShortFileName, VolumeIdx};
 
 type Error = embedded_sdmmc::Error<std::io::Error>;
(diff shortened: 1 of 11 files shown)


eldruin (commented):
Looks good to me, thank you!
Could you add an entry to the changelog about this?

thejpster:
ugh, for goodness sake. I should have a git pre-commit hook or something that warns me about that so you don't have to :/

thejpster:
Looking again at this, I think the last CHANGELOG item already captures the necessary info. Readers of the CHANGELOG need to see the major differences between versions, not the churn that happened within one release.

The entry is:

> - Updated directory iterator callback, to allow bailing out early (breaking change)

eldruin (approved):
I see. Thank you!

PR #221, opened 2026-02-17
Fix 'FAT12 is unsupported' error when opening small FAT32 partitions.

leighleighleigh:
I ran into this misleading error message when using very small in-memory FAT32 filesystems.
These changes fixed it right up. I haven't tested it on anything other than FAT32.

### DIFF
--- a/src/fat/bpb.rs
+++ b/src/fat/bpb.rs
@@ -3,8 +3,13 @@
 use crate::{
     blockdevice::BlockCount,
     fat::{FatType, OnDiskDirEntry},
+    PARTITION_ID_FAT16,PARTITION_ID_FAT16_LBA,PARTITION_ID_FAT16_SMALL,PARTITION_ID_FAT32_CHS_LBA,PARTITION_ID_FAT32_LBA,
 };
 use byteorder::{ByteOrder, LittleEndian};
+
+const PARTITION_INFO_START : usize = 446;
+const PARTITION_INFO_TYPE_INDEX : usize = 4;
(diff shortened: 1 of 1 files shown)


thejpster:
The algorithm used here to detect the type of FAT is exactly as specified by Microsoft in https://academy.cba.mit.edu/classes/networking_communications/SD/FAT.pdf.

How have you ended up with a non-compliant FAT32 volume?

leighleighleigh:
Thank you for the link - I clearly didn't do my homework, my apologies.

For what it's worth, I've been making disk images like so - then embedding them into firmware, and accessing them over USB MSC. This has resulted in a simple cross-platform method to configure the embedded device.

```shell
# filename of the output disk
disk="disk.img"
# directory containing files to put on the disk
seed="./files"
# set the disk size in blocks
block_count=224
# check the seed folder exists
test -d "${seed}"
if [[ ! $? -eq 0 ]];
then
  echo "Seed directory ${seed} not found." >&2
  exit 1
fi
# make a blank disk 
dd if=/dev/zero of="${disk}" bs=512 count=$block_count
# format it as FAT32 with an MBR partition
mkfs.fat -F32 --mbr=y "${disk}"
# use 'mcopy' to copy each file into the disk.img
for f in "${seed}"/*;
do
  mcopy -i "${disk}" $f ::
  if [[ ! $? -eq 0 ]];
  then
    echo "$f - Failed"
  else
    echo "$f - OK"
  fi
done
```

thejpster:
Ah, neat. I guess the issue is that you forced FAT32. I’d be happy to take patches that implement FAT12 if you still end up with floppy disk sized images.

PR #224, opened 2026-02-23
Restructuring, CID support

robamu:
I restructured the low level SD modules a bit.

There is a new `spi` module for the SPI wrapper. I also added a Rust abstraction for the CID register and put the CSD abstractions into a `csd` module


robamu:
okay, some test for the new CID module with the CID I retrieved might be useful as well.

thejpster (approved):

PR #228, opened 2026-02-27
use enums instead of constants for CMD and ACMD

robamu:
what do you think about using enumerations instead of constants?

### DIFF
--- a/src/sdcard/mod.rs
+++ b/src/sdcard/mod.rs
@@ -63,35 +63,35 @@ pub const ERROR_OK: u8 = 0x00;
 #[repr(u8)]
 pub enum CmdId {
     /// GO_IDLE_STATE - Send cards to IDLE state
-    _0_GoIdleState = 0,
+    CMD0_GoIdleState = 0,
     /// ALL_SEND_CID - Request Card IDentification (CID)
-    _2_AllSendCid = 2,
+    CMD2_AllSendCid = 2,
     /// SEND_RELATIVE_ADDR - Request relative card address (RCA)
-    _3_SendRelativeAddr = 3,
+    CMD3_SendRelativeAddr = 3,
     /// SEND_IF_COND - verify SD Memory Card interface operating condition
-    _8_SendIfCond = 8,
+    CMD8_SendIfCond = 8,
     /// SELECT/DESELECT_CARD - Select the active card or deselect the active card
-    _7_SelectCard = 7,
+    CMD7_SelectCard = 7,
     /// SEND_CSD - read the Card Specific Data (CSD register)
-    _9_SendCsd = 9,
+    CMD9_SendCsd = 9,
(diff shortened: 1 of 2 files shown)


src/sdcard/spi.rs
```diff
@@ -209,16 +209,16 @@ where
 
         if blocks.len() == 1 {
             // Start a single-block read
-            self.card_command(CMD17, start_idx)?;
+            self.card_command(super::CmdId::_17_ReadSingleBlock as u8, start_idx)?;
```
thejpster: Shouldn't the `card_command` function take values of the enum type rather than raw `u8` values? Then we wouldn't need the casts everywhere.

src/sdcard/mod.rs
```diff
@@ -37,43 +37,61 @@ pub mod spi;
...
 
-//==============================================================================
+/// Raw command IDs.
+#[bitbybit::bitenum(u6, exhaustive = false)]
+#[derive(Debug, PartialEq, Eq)]
+#[allow(non_camel_case_types)]
+#[repr(u8)]
+pub enum CmdId {
+    /// GO_IDLE_STATE - Send cards to IDLE state
+    _0_GoIdleState = 0,
```
thejpster: These might look better with the digits at the end, like `GoIdleState_0`? Not sure.

src/sdcard/spi.rs
robamu: that is what I though as well. but then the enum should be reasonably complete.

src/sdcard/mod.rs
robamu: I like the number prefix more even though it looks a bit weird. another option would be `Cmd<number>Purpose` but then the ` Cmd`  is always duplicated

src/sdcard/spi.rs
thejpster: It’s complete enough for our needs?

robamu: yes, not an issue :) I will update the `card_command`  function

src/sdcard/mod.rs
thejpster: I'm just not a fan of the leading underscore - I try and avoid it where I can.

Options (in my order of preference, best first):

1. `CmdId::CMD7_SelectCard`
1. `CmdId::Cmd7_SelectCard`
1. `CmdId::Cmd7SelectCard`
1. `CmdId::SelectCard_CMD7`
1. `CmdId::SelectCard_7`
1. `CmdId::_7_SelectCard`

From an auto-complete point of view, most people know the number (because they are working from the spec), so putting that first makes sense. Following it with the name helps people reviewing the code who might not have the spec open and/or don't remember what the number is.

Whatever we pick, we should do the same for the ACMDs, naturally.

robamu: done :)

thejpster (approved):

PR #231, opened 2026-03-02
CSD structure now has Copy and Clone derives

robamu:


thejpster (commented):
clone I can get behind, but do we need to make a 20 byte structure Copy? It feels like you wouldn’t want to accidentally have too many of them on your stack.

I’m open to persuasion though.

robamu:
yeah that is fair. I removed the `Copy` implementation. `Clone`  is sufficient.

thejpster (approved):

PR #232, opened 2026-03-02
start adding argument and response module

robamu:
Nice types for SD mode arguments and responses using `bitbybit`

### DIFF
--- a/src/sdcard/argument.rs
+++ b/src/sdcard/argument.rs
@@ -45,7 +45,7 @@ pub enum PowerControl {
     MaximumPerformance = 1,
 }
 
-/// HPC.
+/// Host Capacity Support (HCS).
 #[bitbybit::bitenum(u1, exhaustive = true)]
 #[derive(Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
@@ -125,7 +125,7 @@ pub struct Acmd6 {
     forbid_overlaps
 )]
 pub struct Acmd41 {
-    /// HPC.
+    /// HCS.
     #[bit(30, rw)]
     host_capacity_support: HostCapacitySupport,
     /// FB.
--- a/src/sdcard/argument.rs
+++ b/src/sdcard/argument.rs
@@ -1,6 +1,7 @@
 //! # Argument module
 //!
-//! Arguments are additional command parameters.
+//! Arguments are additional command parameters. You can use the `sd_card_init` example to see
+//! how these data structures can be used during a SD card initialization process.
 
 use arbitrary_int::u24;
 
--- a/src/sdcard/argument.rs
+++ b/src/sdcard/argument.rs
@@ -1,0 +1,179 @@
+//! # Argument module
+//!
+//! Arguments are additional command parameters.
+
+use arbitrary_int::u24;
+
+/// Voltage settings supplied in CMD8.
+#[bitbybit::bitenum(u4, exhaustive = false)]
+#[derive(Debug, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
+pub enum VoltageSuppliedSelect {
+    /// Regular voltage range.
+    #[default]
+    _2_7To3_6V = 0b0001,
+    /// Reserved for low voltage.
+    LowVoltageRange = 0b0010,
+}
+
+/// CMD8 argument.
+#[bitbybit::bitfield(u32, default = 0x0, debug, defmt_fields(feature = "defmt-log"))]
+pub struct Cmd8 {
+    /// PCIe v1.2 support.
+    #[bit(13, rw)]
+    pcie_1_2v_support: bool,
+    /// PCIe available.
+    #[bit(12, rw)]
+    pcie_availability: bool,
+    /// Voltage settings.
+    #[bits(8..=11, rw)]
+    voltage_supplied: Option<VoltageSuppliedSelect>,
+    /// Check pattern which is echoed.
+    #[bits(0..=7, rw)]
+    check_pattern: u8,
+}
+
+/// Power control (XPC) settings.
+#[bitbybit::bitenum(u1, exhaustive = true)]
+#[derive(Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
+pub enum PowerControl {
+    /// Power saving.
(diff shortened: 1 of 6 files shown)


src/sdcard/mod.rs
```diff
@@ -77,6 +77,378 @@ pub const ACMD41: u8 = 0x29;
 
 //==============================================================================
 
+/// Argument module. Arguments are additional command parameters.
+pub mod argument {
```
thejpster: I’d rather have modules on disk instead of inline modules. I just it makes them easier to find in your editor?

robamu: I can do that

thejpster:
Looks good, but I see these types aren't yet used. Is there a plan to use them?

robamu:
for now, these can be used for implementing SD drivers in separate crates. Implementing a generic SD driver which use these inside the library could work, but a generic transport trait to send CMDs/ACMDS is required first and I am not fully sure how well that works.

robamu:
rebased again

src/sdcard/argument.rs
```diff
@@ -0,0 +1,179 @@
...
+#[derive(Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt-log", derive(defmt::Format))]
+pub enum PowerControl {
+    /// Power saving.
+    PowerSaving = 0,
+    /// Maximum performance.
+    MaximumPerformance = 1,
+}
+
+/// HPC.
```
thejpster, line 47: HPC?

thejpster:
Looks ok, but I think we need some kind of example to show how to use them. How hard would it be to build a fake SD Card with an API to send requests and get responses that an example can drive in a similar way to how you would initialise an actual SD Card?

robamu:
hmm. probably okay? For most responses, it is relatively clear what the SD card needs to send back, or at the very least what is relevant. I can look into this. If we do something like that, we should write to so it can be re-used for tests.

robamu:
I added a `SdCardMock` which is basically a virtualized SD card. currently, I only implemented the internal state machine to allow performing the SD card initialization with it.

robamu:
In the SD card example, we could still implement the primary SD driver traits.. this makes it a bit clearer how everything fits together I suppose.

robamu:
The initialization sequence I implemented is really similar to the ones I implemented for the SAM9G20 SD controller and the Z7 SD controller. the only thing i strongly simplified is that I do not handle the case of pre V2 SD cards where there is not CMD8 reply. for that, I would have to add some sort of way to simulate CMD timeouts.

src/sdcard/mock.rs
```diff
@@ -0,0 +1,267 @@
...
+/// The IDLE mode has some private substates used for operating condition negotiation.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum IdleSubState {
+    /// Idle substate.
+    Idle,
+    /// Received CMD8, ready to accept ACMD41.
+    ReceivedIfCond,
+    /// Received ACMD41 with OCR bit, initializing itself now.
+    ///
+    /// This allows for the intialization to take multiple polling calls.
```
thejpster:
```suggestion
    /// This allows for the initialization to take multiple polling calls.
```

src/sdcard/mock.rs
```diff
@@ -0,0 +1,267 @@
...
+    InitializingSelf {
+        /// Step counter. When a treshold is reached, initialization is complete.
```
thejpster:
```suggestion
        /// Step counter. When a threshold is reached, initialization is complete.
```

src/sdcard/mock.rs
```diff
@@ -0,0 +1,267 @@
...
+    AcmdId,
+    argument::{Acmd6, BusWidth, Cmd7, Cmd8, Cmd9, Cmd13},
+    response::{R1, R3, R6, R7},
+};
+
+/// CID retreived from a real SD card, will be returned by the mock object.
+pub const TEST_CID: [u8; 16] = [
+    0x12, 0x34, 0x56, 0x41, 0x53, 0x54, 0x43, 0x0, 0x20, 0x0, 0x0, 0xc, 0xef, 0x1, 0x65, 0xef,
+];
+/// CSD retreived from a real SD card, will be returned by the mock object.
```
thejpster:
```suggestion
/// CSD retrieved from a real SD card, will be returned by the mock object.
```

src/sdcard/mock.rs
```diff
@@ -0,0 +1,267 @@
...
+//! The `sd_card_init` example application shows how the mock is used to initialize a SD card.
+//! This can serve as a reference for implementing this for real devices.
+
+use crate::sdcard::{
+    AcmdId,
+    argument::{Acmd6, BusWidth, Cmd7, Cmd8, Cmd9, Cmd13},
+    response::{R1, R3, R6, R7},
+};
+
+/// CID retreived from a real SD card, will be returned by the mock object.
```
thejpster:
```suggestion
/// CID retrieved from a real SD card, will be returned by the mock object.
```

src/sdcard/mock.rs
```diff
@@ -0,0 +1,267 @@
...
+    ReceivedIfCond,
+    /// Received ACMD41 with OCR bit, initializing itself now.
+    ///
+    /// This allows for the intialization to take multiple polling calls.
+    InitializingSelf {
+        /// Step counter. When a treshold is reached, initialization is complete.
+        step: u8,
+    },
+}
+/// SD card mock states.
```
thejpster, line 37:
```suggestion

/// SD card mock states.
```

src/sdcard/mock.rs
```diff
@@ -0,0 +1,267 @@
...
+            cid: [0; 16],
+            bus_width: BusWidth::_1bit,
+            polling_calls_card_init: 3,
+            reply_buf: [0; 16],
+        }
+    }
+
+    /// Insert a [super::AcmdId] command into the SD card.
+    ///
+    /// Normall, this would involve sending a [super::CmdId::CMD55_AppCmd] first, but this
```
thejpster:
```suggestion
    /// Normally, this would involve sending a [super::CmdId::CMD55_AppCmd] first, but this
```

robamu:
I am onto adding support for alternative mockinng modes (SD1, SD2, SDHC) :)

thejpster (approved):

thejpster:
Looks comprehensive, thanks!

PR #233, opened 2026-03-02
Fix `has_open_handles()` boolean logic

hansmrtn:
`has_open_handles()` currently uses:

```
!(open_dirs.is_empty() || open_files.is_empty())
```

By De Morgan's law this is `!dirs_empty && !files_empty`, so it only returns
`true` when both collections are non-empty. If you have an open directory
but no open files (or vice-versa), it incorrectly returns `false`.

Fixed by changing `||` to `&&`:

```
!(open_dirs.is_empty() && open_files.is_empty())
```
I've included a regression test.


thejpster (approved):
Oops. LGTM. Thank you!

PR #234, opened 2026-03-05
Fix FAT32 directory iteration not stopping on ControlFlow::Break

hansmrtn:
`iterate_fat32` use a plain `break` when the callback returns `ControlFlow::Break(())` or when `is_end()` is hit. This only exits the innermost loop and the iteration continues with the next block and next cluster calling the callback with more entries. I believe in practice this still returns the correct result, but the callback is invoked more times than it
should be so unnecessary block I/O is performed.

The equivalent `iterate_fat16` uses `break 'outer` to exit all three nested loops.

Fixed by adding an `'outer:` label to the `while` loop and changing both
`break` statements to `break 'outer`, matching `iterate_fat16`.

The pr includes a regression test to validate this.

### DIFF
--- a/src/fat/volume.rs
+++ b/src/fat/volume.rs
@@ -761,7 +761,7 @@ impl FatVolume {
             ClusterId::ROOT_DIR => Some(fat32_info.first_root_dir_cluster),
             _ => Some(dir_info.cluster),
         };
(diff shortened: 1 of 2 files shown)


thejpster (approved):

thejpster:
10/10. No Notes.

PR #235, opened 2026-03-18
Fix lifetime of returned dierctories and files

Conaclos:
This Pr fixes the lifetime of `open_dir`, `open_root_dir` and `open_file_in_dir` by using the lifetime `'a` of `Directory<'a>` instead of relying on the temporary lifetime of `&self`.

The change is backward compatible.

Note that we could simplify the return type by simply using `Self`. I am not sure if this adheres to the code style of this project.

### DIFF
--- a/src/filesystem/directory.rs
+++ b/src/filesystem/directory.rs
@@ -120,7 +120,7 @@ where
     pub fn open_dir<N>(
         &self,
         name: N,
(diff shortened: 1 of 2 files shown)


thejpster (approved):
Good catch, thanks!

thejpster:
If you wanted to clean up return types using Self, I’d take that as a PR.

PR #236, opened 2026-03-19
IUse Self where it makes the return type easier to read/parse

Conaclos:
As [suggested here](https://github.com/rust-embedded-community/embedded-sdmmc-rs/pull/235#issuecomment-4088389942) I am opening this PR to simplify the return type of some function using `Self`. This makes it easier to parse the type and to understand what is returned.

Note that there exists a Clippy rule `use_self`, however it forces using `Self` everywhere where applicable.

### DIFF
--- a/src/blockdevice.rs
+++ b/src/blockdevice.rs
@@ -101,7 +101,7 @@ where
     D: BlockDevice,
 {
     /// Create a new block cache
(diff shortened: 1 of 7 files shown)


thejpster (approved):

thejpster:
Thank you!

PR #239, opened 2026-03-26
CSD unchecked constructor

robamu:
I needed this on the Zynq7000, where the CRC byte is just missing from the byte stream for some reason.

### DIFF
--- a/src/sdcard/csd.rs
+++ b/src/sdcard/csd.rs
@@ -547,3 +547,668 @@ impl CsdV3 {
         self.crc().value() == calculated
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_csdv1b() {
+        const EXAMPLE_HEX: [u8; 16] = hex!("00 26 00 32 5F 59 83 C8 AD DB CF FF D2 40 40 A5");
+        const EXAMPLE_U128: u128 = u128::from_be_bytes(EXAMPLE_HEX);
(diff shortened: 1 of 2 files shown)


src/sdcard/csd.rs
```diff
@@ -29,20 +35,24 @@ pub enum CsdCreationError {
...
     pub fn new(raw: &[u8; 16]) -> Result<Csd, CsdCreationError> {
+        let unchecked = Self::new_unchecked(raw)?;
+        if !unchecked.verify_crc7() {
+            return Err(CsdCreationError::Checksum);
+        }
+        Ok(unchecked)
+    }
+
+    /// Construct a [Csd] without verifying the checksum.
+    pub fn new_unchecked(raw: &[u8; 16]) -> Result<Self, InvalidCsdStructureFieldError> {
```
thejpster, line 52: I’m not sure the extra error type is helping. I’d just use the existing error type in the Result here and note in the documentation which error variants can actually be generated in this function.

robamu: One thing it is useful for is the auto conversion. That can also be implemented manually though. I might do it like that.

robamu: Oh, read that too fast. I actually like this single error types because they make it clear at the type level what can go wrong. No need to document error variants either then, and you still have an easy conversion to the unified type with the `From` impl.

robamu: Would you still prefer a unified error type?

thejpster: Nah, you persuaded me.

thejpster (approved):

PR #245, opened 2026-05-09
new embedded-sdmmc-types crate

robamu:
- Extract fundamental traits and types into separate crate
- HALs or other libraries which just want to provide `embedded-sdmmc`
  support can include this library instead of the full `embedded-sdmmc`
  library to implement driver support. I would expect `embedded-sdmmc-types` to change less
  frequently than the `embedded-sdmmc` crate using it. It allows `embedded-sdmmc` to evolve
  without breaking the fundamental traits.
- Extended `sd_card_init` example to also implement the block device
  trait. It also only depends on `embedded-sdmmc-types` types now.


robamu:
we could do a full re-export of the types trait in `embedded-sdmmc`

robamu:
I also updated the README now. the top-level README has a mono-repo style short description with links to the contained libraries. the previous documentation was mostly moved to `embedded-sdmmc/README.md`

embedded-sdmmc-types/src/sdcard/mock.rs
```diff
@@ -94,7 +94,8 @@ impl From<State> for super::response::State {
 
 /// SD card mock.
 ///
-/// This basically behaves like a virtual SD card. Currently, this mock simulates a Ver2.00 or later
+/// This basically behaves like a virtual SD card controller with a connected virtual SD card.s
```
thejpster:
```suggestion
/// This basically behaves like a virtual SD card controller with a connected virtual SD card.
```

embedded-sdmmc-types/README.md
```diff
@@ -0,0 +1,35 @@
+# Embedded SD/MMC [![crates.io](https://img.shields.io/crates/v/embedded-sdmmc-types.svg)](https://crates.io/crates/embedded-sdmmc-types) [![Documentation](https://docs.rs/embedded-sdmmc-types/badge.svg)](https://docs.rs/embedded-sdmmc-types)
+
+Embedded SD/MMC Types Library
```
thejpster: This file has two headings (a `===` underline works like a `#` I think) which looks odd?

robamu: ah that was a mistake

robamu:
@thejpster if it helps, i can also help out with maintenance of the crate, including releasing new crates :)

thejpster:
> @thejpster if it helps, i can also help out with maintenance of the crate, including releasing new crates :)

For that I think you need to open a PR on https://github.com/rust-embedded-community/meta and get the majority to approve.

robamu:
yay, I was added to the org :) Do you still have concerns about the updated structure?

thejpster:
I *think* I would move the `BlockCache` back into the `embedded-sdmmc` library, as it's an internal implementation detail that is probably going to evolve and update. I don't think it needs to be exported in a stable way for HALs to use - they can write their own block cache.

But apart from that I think this is OK.

robamu:
okay, I moved it back into embedded-sdmmc

thejpster (approved):

PR #248, opened 2026-06-02
Decode CSD based on CSD_STRUCTURE, not card_type

fdomke:
The type of the CSD depends on the `CSD_STRUCTURE` field, not the physical layer of the card.

Specifically, there are Standard Capacity SD cards that still support CMD8 (`SD2`); 
`csd::Csd::new` already handles this correctly by looking at `CSD_STRUCTURE` instead.


thejpster (approved):
Makes sense to me. Thanks!

thejpster:
Have you tested this on hardware? I’m not in a position to do that myself this week.

